PR #9, opened 2021-09-14
Add early dynamic resource loading

Tadinu:
## Overview
Under a Project/Plugin's Content folder, there would be:
* Dynamic Contents: assets including static meshes, materials, textures, being referenced during Sim run.
* Static Contents: assets statically referenced and configured by actor, material, etc. BPs.
Both of these are both baked when the package is built and loaded up at run time.

For dynamic contents, early async resource loading should be employed at Sim's initialization, so during Sim's run, they could be used instantly, without having to invoke `StaticLoadObject/LoadObject` which requires passing a hard-coded path to the asset and thus compromises the performance. (Ref: void [AWorldLoader::ConstructLevel()](https://github.com/rapyuta-robotics/io_amr_UE/blob/master/Source/io_amr_UE/WorldLoader.cpp#L105))

Reference:
"Avoid Referencing Assets by String"
https://docs.unrealengine.com/en-US/Resources/SampleGames/ARPG/BalancingBlueprintAndCPP/index.html

## Implementation Details

- [x] Add RRGameSingleton, holding loaded resource assets data (static meshes, materials, textures, etc.)
- [x] Add RRGeneralUtils, RRAssetUtils, RRTypeUtils under Tools folder.
- [x] RRObjectCommon, holding common structs like `ERRResourceDataType`, `FRRResource, FRRResourceInfo` which could be included globally to avoid cyclic inclusion among header files.
- [x] **[RapyutaSimulationPlugins.Build.cs]** Add `PhysicsCore` dependency, for linking `UPhysicalMaterial`, a particular type of material resource.
- [x] **[Content]** Add DynamicContents + temporary RapyutaGeneric map for testing this dynamic resource loading
- [x] Add `RRGameMode`, which initializes early resource loading before other GameState, PC's BeginPlay
- [x] Add `LogRapyutaCore` category for the plugin

## How to test
In the application project,
1- Configure Level map's Override Gamemode as `RRGameMode`
2- In [DefaultEngineBase.ini](https://github.com/rapyuta-robotics/io_amr_UE/blob/master/Config/DefaultEngineBase.ini)
Under [/Script/Engine.Engine], add:
`GameSingletonClassName=/Script/RapyutaSimulationPlugins.RRGameSingleton`
Load the map up, there should appear 2 messages on screen:
```
LOADING SIM DYNAMIC RESOURCES..
ALL DYNAMIC RESOURCES LOADED
```

## Note:
* The specifc asset object (`UStaticMesh*, UMaterialInterface*, etc.)` would then be fetched through `URRGameSingleton::GetStaticMesh(InStaticMeshName) & ::GetMaterial(InMaterialName)` respectively
* Since the resources are stored in a TMap with its name as key, each static mesh or material is expected to have unique asset name of its type, regardless of which folder it resides.
* This PR includes the early loading for **StaticMesh** and **Material** only. For Texture, Curve, etc., they will be added in other PRs upon specific requirements of application projects.

## Risk
* Without steps being taken above, this PR does not impact the stakeholder projects (io_amr_UE/turtlebot3_UE) yet.

## Next step
- Update `ROS2GameMode` BP: made to inherit `RRGameMode` in case some client map that is configured as `ROS2GameMode` uses dynamic resources.
- Update `io_amr_UE/turtlebot3_UE` following above steps & using URRGameSingleton's resource fetching API


cconti-rr (commented):
is there a use for RapyutaGeneric.umap?
seems to be empty
if there's no use, please remove it

do we also need to have copies of basic shape in this plugin for this to work?

tetorea (commented):
Testing this in turtlebot project, I get following warnings and error
```
LogLinker: Warning: Failed to load '/Script/UE_rapyuta_assets': Can't find file.
LogLinker: Warning: Failed to load '/Script/UE_rapyuta_assets': Can't find file.
LogUObjectGlobals: Warning: Failed to find object 'Class /Script/UE_rapyuta_assets.RRGameSingleton'
LogEngine: Error: Engine config value GameSingletonClassName '/Script/UE_rapyuta_assets.RRGameSingleton' is not a valid class name.
```
my forlder turtlebot3-UE/Script/ is empty actually

Tadinu:
> is there a use for RapyutaGeneric.umap?
> seems to be empty
> if there's no use, please remove it
> 
> do we also need to have copies of basic shape in this plugin for this to work?

Yeah, the map is only for testing this PR. It would be removed then.
Regarding the basic shapes or dynamic resources in general, so 
* They should be moved to the specific folder `DynamicContents`, which Sim will load from. 
* For other static assets that are not referenced during Sim run, they could be kept as they are, under our project/plugin's Content outside of (`DynamicContents`) or built-in Engine's Content folder.

@tetorea Sorry, I just updated the description on Steps of `How to Test` to our new plugin name, which should be: 
```
GameSingletonClassName=/Script/RapyutaSimulationPlugins.RRGameSingleton
```
The config path in `.INI` file is actually not a physical path on disk, and only used as the internal key to a `UPROPERTY(config)` member value.

cconti-rr (approved):

tetorea:
@Tadinu thanks, it's better with the new line
I get the messages when I play the RapyutaGeneric map. The environment is completely black though, is it normal?

yuokamoto (approved):

PR #12, opened 2021-09-22
avoid mesh loading from cpp

yuokamoto:


Tadinu (approved):

chalice-graeme (approved):
Works fine with a packaged version of turtlebot3-UE. LGTM.

PR #16, opened 2021-10-08
Odom noise

cconti-rr:
This feature has only been tested within UE4 with the turtlebot3-UE repo and not interfaced with rviz/other ros software.
As publication of the odometry topic worked before (in turtlebot3-UE), it is assumed that it still does.

A series of UE_LOG statements can be uncommented to do some verification.

Ideally it should be tested within the io_amr_ue infrastructure.

This PR should not affect other parts of the code.

PS a 3d lidar BP has been added to this branch by mistake - as it is just a BP and there's nothing depending on it, it should be safe.

### DIFF
--- a/Source/RapyutaSimulationPlugins/Private/Sensors/BaseLidar.cpp
+++ b/Source/RapyutaSimulationPlugins/Private/Sensors/BaseLidar.cpp
@@ -8,8 +8,6 @@
 ABaseLidar::ABaseLidar()
 {
     Gen = std::mt19937{Rng()};
-    GaussianRNGPosition = std::normal_distribution<>{PositionalNoiseMean, PositionalNoiseVariance};
-    GaussianRNGIntensity = std::normal_distribution<>{IntensityNoiseMean, IntensityNoiseVariance};
 
     PrimaryActorTick.bCanEverTick = true;
 
@@ -22,6 +20,8 @@ void ABaseLidar::BeginPlay()
 {
 	Super::BeginPlay();
 	
+    GaussianRNGPosition = std::normal_distribution<>{PositionalNoiseMean, PositionalNoiseVariance};
+    GaussianRNGIntensity = std::normal_distribution<>{IntensityNoiseMean, IntensityNoiseVariance};
 }
 
 // Called every frame


Source/RapyutaSimulationPlugins/Private/Drives/RobotVehicleMovementComponent.cpp
```diff
@@ -59,22 +74,52 @@ void URobotVehicleMovementComponent::UpdateOdom(float DeltaTime)
...
     OdomData.pose_pose_position_z = Pos.Z;
-    OdomData.pose_pose_orientation = FQuat(PawnOwner->GetActorRotation() - InitialTransform.GetRotation().Rotator());
-
-    // velocity
-    OdomData.twist_twist_linear = Velocity;
-    OdomData.twist_twist_angular = FMath::DegreesToRadians(AngularVelocity);
+    OdomData.pose_pose_orientation = Rot;
+
+    // velocity - are signs correct?
+    OdomData.twist_twist_linear = (Pos - PreviousEstimatedPos)/DeltaTime;
```
tetorea, line 107: For linear velocity, in my code, I took into account the orientation modification, something like this:
`FVector NewVelocity = OdomData.pose_pose_orientation.UnrotateVector( Pos-PreviousEstimatedPos ) / DeltaTime;`
so that I can roughly get back the linear and angular velocities that were set with some ros twist msg.
But I don't know if this is the expected way to compute linear velocity for odom msg.

cconti-rr: changed according to your suggestion

it basically depends on the frame of reference that the twist_linear component should be expressed in.
Mine was in global coordinates, while what you suggest is in the frame of reference of the agent, so it should be mostly u=U, y=0, z=0 (where U is the cmd_vel velocity) for vehicles that use differential driving

@yuokamoto what frame of reference is needed?

cconti-rr: I'm adding the gazebo version for the differential drive, while leaving the other as default

cconti-rr:
> nav seems not moves properly. maybe default value toomuch?

yeah, I'm currently trying to solve the problem with Gazebo's odom - I think I fixed the speed (not pushed yet), but there's something wrong with rotation

cconti-rr:
@yuokamoto I added noise, please check
current default noise level is quite low, but it's visible in rviz if you leave the actor still
for the differential drive, I'm only adding to the vr/vl quantities, after linear and angular components are computed, so I don't use 2 RNGs.

@Tadinu please check too

yuokamoto (approved):

Tadinu (approved):

PR #21, opened 2021-11-09
Amr spawn with params

larryng-rapyuta:
changed entity spawn service such that spawn is deferred and a spawn parameters object is attached to the entity before its spawned, additionally sets the name for any entity spawned


yuokamoto (approved):

Source/RapyutaSimulationPlugins/Private/Tools/ROS2Spawnable.cpp
```diff
@@ -0,0 +1,42 @@
+//
+// Created by lng on 2021/10/13.
+//
```
Tadinu, line 3: @larryng-rapyuta Copyright note here should be added also?

Source/RapyutaSimulationPlugins/Private/Tools/ROS2Spawnable.cpp
```diff
@@ -0,0 +1,42 @@
...
+    try{
+        SetName(Request.state_name);
+    } catch (const std::exception& e){
+        UE_LOG(LogTemp, Warning, TEXT("No state_name"));
+    }
+    try{
+        SetNamespace(Request.robot_namespace);
+    } catch (const std::exception& e){
+        UE_LOG(LogTemp, Warning, TEXT("No robot_namespace"));
+    }
```
Tadinu, line 21: * We now have `LogRapyutaCore` as defined [here](https://github.com/rapyuta-robotics/RapyutaSimulationPlugins/blob/main/Source/RapyutaSimulationPlugins/Private/RapyutaSimulationPlugins.cpp#L19), so it could be used I guess.
* Since both of the name setting functions look quite straightforward, not sure which exceptions could happen tho?

Source/RapyutaSimulationPlugins/Private/Tools/SimulationState.cpp
```diff
@@ -256,22 +256,35 @@ void ASimulationState::SpawnEntitySrv(UROS2GenericSrv* Service)
 
             FActorSpawnParameters SpawnParameters;
             SpawnParameters.Name = FName(Request.state_name);
+
             FRotator Rotator = Request.state_pose_orientation.Rotator();
             FVector Position(Request.state_pose_position_x, Request.state_pose_position_y, Request.state_pose_position_z);
-            AActor* NewEntity = GetWorld()->SpawnActor(SpawnableEntities[Request.xml], &Position, &Rotator, SpawnParameters);
+            FVector Scale(1, 1, 1);
+            FTransform Transform(Rotator, Position, Scale);
```
Tadinu, line 263: We could use Scale as `FVector::OneVector` but I suppose its value is defaulted in `FTransform` ctor.

Source/RapyutaSimulationPlugins/Public/Tools/ROS2Spawnable.h
```diff
@@ -0,0 +1,43 @@
+//
+// Created by lng on 2021/10/13.
+//
```
Tadinu, line 3: Copyright note I guess.

Tadinu (approved):
This looks really nice overall, thanks!

Source/RapyutaSimulationPlugins/Private/Tools/ROS2Spawnable.cpp
```diff
@@ -0,0 +1,42 @@
...
+
+void UROS2Spawnable::SetName(FString Name)
+{
+    ActorName = Name;
+}
+
+void UROS2Spawnable::SetNamespace(FString Namespace)
+{
+    ActorNamespace = Namespace;
+}
```
Tadinu, line 32: The parameter could be of `const FString&`, and if we could check for its emptiness as you might want (as in above exceptions), we could log the warnings here-in I suppose.

Source/RapyutaSimulationPlugins/Private/Tools/ROS2Spawnable.cpp
larryng-rapyuta: Oh wasn't used to that, I'll look and see what the other files do with the copyright message and copy it over

Source/RapyutaSimulationPlugins/Private/Tools/ROS2Spawnable.cpp
larryng-rapyuta: Ahh i see, ill take a look at the LogRapyutaCore. I set these exceptions thinking that there maybe a case in the future where a changed spawnEntity message or the entity being spawned might not need to use all the parameters, and having these parameters being set as null might cause issues

Source/RapyutaSimulationPlugins/Private/Tools/SimulationState.cpp
larryng-rapyuta: ohh Ill try that instead!

yuokamoto (approved):

PR #24, opened 2021-12-12
add ROS2 rgb camera plugin

yuokamoto:


Source/RapyutaSimulationPlugins/Private/Sensors/ROS2Camera.cpp
```diff
@@ -0,0 +1,77 @@
...
+
+FROSImage AROS2Camera::GetData()
+{
+    SceneCaptureComponent->CaptureScene();
+
+    FTextureRenderTarget2DResource* RenderTargetResource;
+    RenderTargetResource = (FTextureRenderTarget2DResource*)RenderTarget->GameThread_GetRenderTargetResource();
+    if (RenderTargetResource) {
+        TArray<FColor> buffer;
+        RenderTargetResource->ReadPixels(buffer);
```
Tadinu, line 67: Just a note, this would be blocking if called in high frequency. A non-blocking version could be referenced [here](https://github.com/TimmHess/UnrealImageCapture).
Some discussions on a related feature being implemented in AirSim:
https://github.com/microsoft/AirSim/pull/3018
https://github.com/Microsoft/AirSim/issues/1818
https://github.com/microsoft/AirSim/blob/master/Examples/DataCollection/StereoImageGenerator.hpp

Other related:
https://github.com/ayumax/WindowCapture2D
https://github.com/rnixik/StereoPano
https://blog.jamie.holdings/2019/09/08/taking-screenshots-in-game-in-unreal-engine-4-23/

tetorea (approved):

Source/RapyutaSimulationPlugins/Private/Sensors/ROS2CameraActor.cpp
```diff
@@ -0,0 +1,20 @@
+// Copyright 2020-2021 Rapyuta Robotics Co., Ltd.
+
+#include "Sensors/ROS2CameraActor.h"
+
+AROS2CameraActor::AROS2CameraActor() : Super()
```
Tadinu, line 5: `Super()` would be already automatically called, thus is not required for a default constructor.

Source/RapyutaSimulationPlugins/Private/Sensors/ROS2CameraComponent.cpp
```diff
@@ -0,0 +1,144 @@
...
+    // Initialize image data
+    Data.width = Width;
+    Data.height = Height;
+    Data.encoding = Encoding;
+    Data.step = Width * 3; // todo should be variable based on encoding
+    Data.data.AddUninitialized(Width * Height * 3);
+
+    // Node and publisher initialize
+    FActorSpawnParameters SpawnParamsNode;
+    Node = GetWorld()->SpawnActor<AROS2Node>(AROS2Node::StaticClass(), SpawnParamsNode);
```
Tadinu, line 38: If we are using a default `FActorSpawnParameters`, maybe no need to pass it in, as well as the `AROS2Node::StaticClass()`, which is already deduced from the template type as `AROS2Node`:
```suggestion
   Node = GetWorld()->SpawnActor<AROS2Node>();
```

Source/RapyutaSimulationPlugins/Private/Sensors/ROS2CameraComponent.cpp
```diff
@@ -0,0 +1,144 @@
...
+    Node->Name = NodeName.IsEmpty() ? TEXT("UE4CameraNode_" + FGuid::NewGuid().ToString()) : NodeName;
```
Tadinu, line 39: Just a suggestion, normally I used owner/parent comp's name, which is already made sure to be assigned uniquely by UE, as the affix:
```suggestion
    Node->Name = NodeName.IsEmpty() ? FString::Printf(TEXT("ROS2CameraNode_%s"), *GetName()) : NodeName;
```

Source/RapyutaSimulationPlugins/Private/Sensors/ROS2CameraComponent.cpp
```diff
@@ -0,0 +1,144 @@
...
+    Publisher->UpdateDelegate.BindDynamic(this, &UROS2CameraComponent::MessageUpdate);
+    Node->AddPublisher(Publisher);
+    Publisher->Init(UROS2QoS::KeepLast);
+}
+
+void UROS2CameraComponent::MessageUpdate(UROS2GenericMsg *TopicMessage)
+{
+    UROS2ImageMsg *Message = Cast<UROS2ImageMsg>(TopicMessage);
+    Message->SetMsg(GetData());
+}
```
Tadinu, line 58: We now have a template for ~ROS2 Publisher classes, like in [UROS2ClockPublisher](https://github.com/rapyuta-robotics/RapyutaSimulationPlugins/blob/main/Source/RapyutaSimulationPlugins/Private/Tools/ROS2ClockPublisher.cpp), so should we have a class for `UROS2ImagePublisher` for consistency, also helping make this setup a bit shorter? then we only need to configure `TopicName` & `PublicationFrequencyHz` here.

Source/RapyutaSimulationPlugins/Private/Sensors/ROS2CameraComponent.cpp
```diff
@@ -0,0 +1,144 @@
...
+                    Data.data[I * 3 + 2] = nextRenderRequest->Image[I].B;
+                    // UE_LOG(LogTemp, Warning, TEXT("AsyncTaskDone"));
+                }
+
+                // Delete the first element from RenderQueue
+                RenderRequestQueue.Pop();
+                delete nextRenderRequest;
+            }
+        }
+    }
```
Tadinu, line 84: Since here we do `CaptureNonBlocking()`, which means the GameThread is non-blocking, and the capturing (pixel reading from RHI) could possibly take some time before RHI returns the data, meaning the fence is complete.
Thus, `if(nextRenderRequest->RenderFence.IsFenceComplete())` might not be instantly satisfied.

But more essentially, an instant data fetching function named `GetData()` could be inevitably blocking, if users want to invoke a non-blocking version, they might want to fetch data at a later time, when it is available from RHI, after invoking `CaptureNonBlocking()`?

References:
* TimmHess actually runs the fence check in a [tick](https://github.com/TimmHess/UnrealImageCapture#:~:text=imageWrapper%2D%3EGetCompressed(0)%3B-,RunAsyncImageSaveTask(ImgData%2C%20fileName)%3B,-%7D%0A%0A%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20//%20Delete%20the%20first) then do async image saving.
* [Another one](https://github.com/rapyuta-robotics/RapyutaSimulationInternal/blob/master/Source/RapyutaSimInternal/Private/RapyutaSceneCapture/RRCaptureUtils.cpp#L174) of async fetching style by off-loading the fence check to a thread.

Tadinu: Since here we do `CaptureNonBlocking()`, which means the GameThread is non-blocking, and the capturing (pixel reading from RHI) could possibly take some time before RHI returns the data.
Thus, `if(nextRenderRequest->RenderFence.IsFenceComplete())` might not be instantly satisfied.

But more essentially, an instant data fetching function named `GetData()` could be inevitably blocking, if users want to invoke a non-blocking version, they might want to fetch data at a later time, when it is available from RHI, after invoking `CaptureNonBlocking()`?

References:
* TimmHess actually runs the fence check in a [tick](https://github.com/TimmHess/UnrealImageCapture#:~:text=imageWrapper%2D%3EGetCompressed(0)%3B-,RunAsyncImageSaveTask(ImgData%2C%20fileName)%3B,-%7D%0A%0A%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20//%20Delete%20the%20first) then do async image saving.
* [Another one](https://github.com/rapyuta-robotics/RapyutaSimulationInternal/blob/master/Source/RapyutaSimInternal/Private/RapyutaSceneCapture/RRCaptureUtils.cpp#L174) of async fetching style by off-loading the fence check to a thread.

Source/RapyutaSimulationPlugins/Private/Sensors/ROS2CameraActor.cpp
```diff
@@ -0,0 +1,20 @@
...
+
+AROS2CameraActor::AROS2CameraActor() : Super()
+{
+	SceneComponent = CreateDefaultSubobject<USceneComponent>(TEXT("SceneComponent"));
+
+	// Make the scene component the root component
+	RootComponent = SceneComponent;
+	
+	// Setup camera defaults
+	CameraComponent = CreateDefaultSubobject<UROS2CameraComponent>(TEXT("CameraComponent"));
```
Tadinu, line 13: Since `UCameraComponent` itself is already of `USceneComponent`, maybe it could become the root, so we would not need to create an extra scene component for it I suppose.

Source/RapyutaSimulationPlugins/Public/Sensors/ROS2CameraComponent.h
```diff
@@ -0,0 +1,87 @@
...
+#pragma once
+
+#include "CoreMinimal.h"
+#include "Camera/CameraComponent.h"
+#include "Components/SceneCaptureComponent2D.h"
+#include "Engine/TextureRenderTarget2D.h"
+
+#include "ROS2Node.h"
+#include <Msgs/ROS2ImageMsg.h>
+#include "ROS2Publisher.h"
```
Tadinu, line 12: Just a note to avoid auto-reordering of header files by `clang-format`, by adding `//UE, //rclUE` on respective groups.

Source/RapyutaSimulationPlugins/Public/Sensors/ROS2CameraComponent.h
```diff
@@ -0,0 +1,87 @@
...
+	UROS2Publisher* Publisher;
+
+	UFUNCTION(BlueprintCallable)
+	virtual FROSImage GetData();
+
+	UPROPERTY(EditAnywhere, BlueprintReadWrite)
+	FString NodeName = TEXT("");
+
+	UPROPERTY(EditAnywhere, BlueprintReadWrite)
+	FString Namespace = TEXT("");
```
Tadinu, line 71: `FString()` already means `TEXT("")`?

Source/RapyutaSimulationPlugins/Public/Sensors/ROS2CameraComponent.h
```diff
@@ -0,0 +1,87 @@
...
+	FString TopicName = TEXT("img_raw");
+
+	UPROPERTY(EditAnywhere, BlueprintReadWrite)
+	int32 PublishFreq = 1;
+	
+	UPROPERTY(EditAnywhere, BlueprintReadWrite)
+	int Width = 640;
+	
+	UPROPERTY(EditAnywhere, BlueprintReadWrite)
+	int Height = 480;
```
Tadinu, line 83: `int -> int32` to make it explicit?

Source/RapyutaSimulationPlugins/Private/Sensors/ROS2CameraComponent.cpp
yuokamoto: it will become /ROS2CameraNode_CameraComponent. I guess component name is not unique.

Tadinu: Each, all actors/components in UE are basically `UObject`, of which the names must be unique, as noted [here](https://github.com/EpicGames/UnrealEngine/blob/release/Engine/Source/Runtime/CoreUObject/Public/UObject/UObjectGlobals.h#L1163), or as in [UObject::Rename()](https://docs.unrealengine.com/4.27/en-US/API/Runtime/CoreUObject/UObject/UObject/Rename/)
-> So `GetName()` returns that object name, versus `GetClassName()` which returns the type name as `CameraComponent`.

But yeah, if we then might run multiple Sim instances, `GUID` would be more reliable I suppose.

Tadinu: All actors/components in UE are basically `UObject`, of which the names must be unique, as noted [here](https://github.com/EpicGames/UnrealEngine/blob/release/Engine/Source/Runtime/CoreUObject/Public/UObject/UObjectGlobals.h#L1163), or as in [UObject::Rename()](https://docs.unrealengine.com/4.27/en-US/API/Runtime/CoreUObject/UObject/UObject/Rename/)
-> So `GetName()` returns that object name, versus `GetClassName()` which returns the type name as `CameraComponent`.

But yeah, if we then might run multiple Sim instances, `GUID` would be more reliable I suppose.

Source/RapyutaSimulationPlugins/Private/Sensors/ROS2CameraComponent.cpp
yuokamoto: I think, it is too much for now since we have only this camera class which publish image. since getdata is tightly coupled with cameracomponent. we can do same as lidar to make publisher class have ROS2CameraComponent, 

we can revist this thing when we add other camera such as depth in the near feature.

Tadinu (approved):

PR #25, opened 2021-12-14
update turtlebot assets

yuokamoto:
* Update TurtlebotBurger mesh assets
* **[ATurtlebotBurger]**: Remove `UDifferentialDriveComponent* DifferentialDriveComponent`, creating it on the fly using the inherited `ARobotVehicle::RobotVehicleMoveComponent`


Tadinu (approved):

Tadinu:
@yuokamoto Since we would not have to re-implement `InitializeMoveComponent()` anymore to create `DifferentialDriveComponent`, could you by the way remove the `virtual` from its [declaration](https://github.com/rapyuta-robotics/RapyutaSimulationPlugins/blob/main/Source/RapyutaSimulationPlugins/Public/Robots/RobotVehicle.h#L21), just in case some child class wants to call it inside a ctor like in [AMRVehicle](https://github.com/rapyuta-robotics/io_amr_UE/blob/master/Source/io_amr_UE/AMRVehicle.cpp#L19).

PR #26, opened 2021-12-16
Add ROS2GameMode in cpp

Tadinu:
## Overview
In preparation of integration Robot importer, which would warrant `ARobotVehicle's RobotVehicleMoveComponent` creation at run-time, and thus also its ros-related information like OdomData, FrameId, ChildFrameId's initialization as well, 
=> There would be certain modifications to the whole initializing process of ROS services, particularly **ROS2Node** & **SimulationState**, which are currently created & initialized by ROS2GameMode in BP.
=> In order to facilitate various custom handling for importing & spawning robots on the fly, a cpp version would be more convenient and also non-user technicalities could be hidden from BP's users.

Thus come this PR with the specific following changes:
- [x] Add ARRROS2GameMode cpp class, inheriting from RRGameMode + Remove ROS2GameMode BP
- [x] Various refactoring, mostly clang-format
+ ATurtlebotROSController::OdomMessageUpdate(): only do the odometry update if OdomData is valid with non-emty frame_id/child_frame_id and covariance values, which is add in [rclUE's PR](https://github.com/rapyuta-robotics/rclUE/pull/26)
+ ATurtlebotROSController::GetOdomData(): update to return bool
- [x] ATurtlebotBurger::Init() Set Base as RootComponent, due to a warning from the UE
- [x] Add cpp `ROS2ClockPublisher` class, for specific initializing setup (frequency + UpdateMessage callback)

There are still some generalization that could be unified for ROS2-TF/Clock/Lidar/String-Publisher, which should be done in another PR.

## Wrike
https://www.wrike.com/open.htm?id=813301215


Source/RapyutaSimulationPlugins/Private/RRROS2GameMode.cpp
```diff
@@ -0,0 +1,35 @@
+// Copyright 2020-2021 Rapyuta Robotics Co., Ltd.
```
yuokamoto, line 1: Can you name it as ROS2GameMode?

Tadinu: Ah sure, I guess since this is a public repo, so we should not prefix class name right?

tetorea (approved):

Source/RapyutaSimulationPlugins/Private/RRROS2GameMode.cpp
```diff
@@ -0,0 +1,35 @@
...
+    ROS2Node = currentWorld->SpawnActor<AROS2Node>();
+    ROS2Node->Namespace.Reset();
+    ROS2Node->Init();
+
+    // Create Clock publisher
+    ClockPublisher = NewObject<UROS2Publisher>(this);
+    ClockPublisher->RegisterComponent();
+    ClockPublisher->TopicName = TEXT("clock");
+    ClockPublisher->PublicationFrequencyHz = 1;
+    ClockPublisher->MsgClass = UROS2StringMsg::StaticClass();
```
yuokamoto, line 27: clock is not stringmsg

yuokamoto (approved):

tetorea (approved):

chalice-graeme (approved):

PR #29, opened 2021-12-24
Add lidar component

Tadinu:
### Overview
In preparation of supporting <gazebo>'s <sensor> subelement in urdf/sdf to specify sensors like lidar, depth, etc, the lidar component would then be created on-the-fly along with other built-up robot components.
Currently, lidar is embodied as an AActor-based class, thus in order to serve that purpose, a SceneComponent-based should be used instead, and here comes this PR mainly refactoring `ABaseLidar, ASensorLidar, ASensorLidar3D` into corresponding lidar component classes.

### Implementation details
- [x] Add `ARRBaseLidarComponent, ARR2DLidarComponent, ARR3DLidarComponent` inheriting from `USceneComponent` wrapping the contents of `ABaseLidar, ASensorLidar, ASensorLidar3D` respectively.
- [x] Add `ARRLidar` inheriting from `AStaticMeshActor` housing lidar component, mostly to be used in case of placing into level if needed.
- [x] **ATurtlebotBurger, ATurtlebotROSController, BP_TurlebotROSController, BP_Turtlebot~, BP_ lidar sensor actors**: Update to using lidar component instead.
- [x] Add cpp **ROS2LidarPublisher**, replacing BP one

https://www.wrike.com/open.htm?id=816995043


yuokamoto (commented):
Why did you remove BP_Turtlebot assets?

Tadinu:
> Why did you remove BP_Turtlebot assets?
I did not, Which specific ones do you mean? Actually, only the lidar actor, which was previously added in BP, is replaced by the lidar component now added in cpp, thus the BP ones technically become the data BP and get configured only without extra BP-added contents.

Tadinu:
> Why did you remove BP_Turtlebot assets?

I did not, Which specific ones do you mean? Actually, only the lidar actor, which was previously added in BP, is replaced by the lidar component now added in cpp, thus the BP ones technically become the data BP and get configured only without extra BP-added contents.

yuokamoto:
> > Why did you remove BP_Turtlebot assets?
> 
> I did not, Which specific ones do you mean? Actually, only the lidar actor, which was previously added in BP, is replaced by the lidar component now added in cpp, thus the BP ones technically become the data BP and get configured only without extra BP-added contents.

Sorry, my bad, I misunderstood

Source/RapyutaSimulationPlugins/Public/Robots/Turtlebot3/TurtlebotBurger.h
```diff
@@ -11,6 +11,7 @@
 
 DECLARE_LOG_CATEGORY_EXTERN(LogTurtlebotBurger, Log, All);
 
+class URR2DLidarComponent;
```
yuokamoto, line 14: please avoid forward declarelation
https://google.github.io/styleguide/cppguide.html#Forward_Declarations

Source/RapyutaSimulationPlugins/Private/Sensors/RRLidar.cpp
```diff
@@ -0,0 +1,13 @@
...
+
+ARRLidar::ARRLidar()
+{
+    LidarComponent =
+        Cast<URRBaseLidarComponent>(CreateDefaultSubobject(TEXT("Lidar"), LidarComponentClass, LidarComponentClass, true, false));
+    UStaticMeshComponent* staticMeshComponent = GetStaticMeshComponent();
+    USceneComponent* rootComponent =
+        staticMeshComponent ? static_cast<USceneComponent*>(staticMeshComponent) : static_cast<USceneComponent*>(LidarComponent);
+    SetRootComponent(rootComponent);
+}
```
yuokamoto, line 13: Don;t need to `SetupAttachment` LidarComponent to rootComponent?

yuokamoto:
https://user-images.githubusercontent.com/16463135/147434762-ff463e3a-1f4b-455e-aaa4-f199f5b96c98.mp4

BP_TurtlebotBurger seems have 2 lidar publishers

Source/RapyutaSimulationPlugins/Public/Robots/Turtlebot3/TurtlebotBurger.h
Tadinu: Right, thanks, kinda my habits to avoid cyclic depedency, which is not really required in this case.

yuokamoto (approved):

tetorea (approved):

PR #31, opened 2022-01-10
Odomupdate

yuokamoto:


Source/RapyutaSimulationPlugins/Public/Robots/RRRobotVehicleROSController.h
```diff
@@ -49,4 +49,10 @@ class RAPYUTASIMULATIONPLUGINS_API ARRRobotVehicleROSController : public AAICont
 
     UFUNCTION(BlueprintCallable)
     void SubscribeToMovementCommandTopic(const FString& InTopicName);
+
+    UPROPERTY(BlueprintReadWrite)
+    bool PublishOdom = true;
+
+    UPROPERTY(BlueprintReadWrite)
+    bool PublishOdomTf = false;
```
Tadinu, line 57: By convention, we could prefix these members with `b` just for quickly recognizing them as boolean. And if making them as `uint8 <bMember> :1` style also (like in various places in UE sources) due to this class being attached to every single robot, we would then have to initialize them in ctor.., which is kinda a catch.

Source/RapyutaSimulationPlugins/Public/Tools/RRROS2OdomPublisher.h
```diff
@@ -40,4 +40,10 @@ class RAPYUTASIMULATIONPLUGINS_API URRROS2OdomPublisher : public UROS2Publisher
     void RevokeUpdateCallback() override;
     void UpdateMessage(UROS2GenericMsg* InMessage) override;
     bool GetOdomData(FROSOdometry& OutOdomData) const;
+
+    UPROPERTY(BlueprintReadWrite)
+    bool PublishOdomTf = false;
+
+    UPROPERTY(BlueprintReadWrite)
+    bool AppendNodeNamespace = true;
```
Tadinu, line 48: Same as above

Source/RapyutaSimulationPlugins/Private/Robots/RRRobotVehicleROSController.cpp
```diff
@@ -80,8 +84,11 @@ void ARRRobotVehicleROSController::OnPossess(APawn* InPawn)
 
 void ARRRobotVehicleROSController::OnUnPossess()
 {
-    OdomPublisher->RevokeUpdateCallback();
-    OdomPublisher->RobotVehicle = nullptr;
+    if (PublishOdom)
+    {
+        OdomPublisher->RevokeUpdateCallback();
+        OdomPublisher->RobotVehicle = nullptr;
+    }
```
Tadinu, line 91: Might we include the check on `OdomPublisher` also for safety?

Source/RapyutaSimulationPlugins/Private/Robots/Turtlebot3/Turtlebot.cpp
```diff
@@ -4,9 +4,10 @@
 
 void ATurtlebot::ConfigureVehicleMoveComponent()
 {
+
     Super::ConfigureVehicleMoveComponent();
     check(RobotVehicleMoveComponent);
     // [FrameIds] will then be used for odom init in Super::InitializeMoveComponent(),
     // where [RobotVehicleMoveComponent's PawnOwner] has been available thus they must be inited beforehands.
-    RobotVehicleMoveComponent->SetFrameIds(TEXT("odom"), TEXT("base_footprint"));
+    // RobotVehicleMoveComponent->SetFrameIds(TEXT("odom"), TEXT("base_footprint"));
```
Tadinu, line 12: We should avoid commenting out code in the long run to save extra refactoring I think. I guess since you have moved these values to their initial default in the class, we could remove this method overriding?
I thought `odom, base_footprint` were very custom values (are they or they are hardcoded in ROS ?), thus initialized them here and did not set them as default values in the generic parent class `ARobotVehicle`, since other child robot types might have different frame id values?

Tadinu (approved):

Source/RapyutaSimulationPlugins/Private/Robots/Turtlebot3/Turtlebot.cpp
yuokamoto: odom base_footprint is very common frame names and many navigation has this as default value as far as I know.
but sometimes use different valus such as base_link

PR #32, opened 2022-01-12
add option to append frame id to lidar

yuokamoto:


Tadinu (approved):
Thanks for the update, just nitpicking, `bAppendNodeNamespace` with prefix `b` by convention.

PR #34, opened 2022-01-14
Add early ros2 start

Tadinu:
- [x] RRROS2GameMode: Move ROS2 start to an earlier moment as in `InitGame()`, which runs before GameState, PC and thus all of its child actors, objects' BeginPlay()
https://rapyuta-robotics.slack.com/archives/C012TP5GKJ4/p1642133937053500?thread_ts=1640311193.370700&cid=C012TP5GKJ4

### DIFF
--- a/Source/RapyutaSimulationPlugins/Private/RRROS2GameMode.cpp
+++ b/Source/RapyutaSimulationPlugins/Private/RRROS2GameMode.cpp
@@ -2,6 +2,9 @@
 
 #include "RRROS2GameMode.h"
 
+// UE
+#include "HAL/PlatformMisc.h"
+
 // rclUE
 #include "Msgs/ROS2ClockMsg.h"
 #include "ROS2Node.h"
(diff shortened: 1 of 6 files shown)


Source/RapyutaSimulationPlugins/Public/Tools/SimulationState.h
```diff
@@ -53,7 +53,7 @@ class RAPYUTASIMULATIONPLUGINS_API ASimulationState : public AActor
 
     // need node that will handle services - this class will only define and register the service
     UPROPERTY(EditAnywhere, BlueprintReadWrite)
-    AROS2Node* ROSServiceNode;
+    AROS2Node* ROSServiceNode = nullptr;
```
chalice-graeme, line 56: Might want to make this property read-only, since it is now assigned in Init().

Source/RapyutaSimulationPlugins/Private/RRROS2GameMode.cpp
```diff
@@ -10,9 +13,39 @@
...
+           FPlatformMisc::NumberOfCores(),
+           FPlatformMisc::NumberOfCoresIncludingHyperthreads(),
+           FPlatformMisc::NumberOfWorkerThreadsToSpawn());
+    UE_LOG(LogRapyutaCore, Display, TEXT("ShouldUseThreadingForPerformance: %d"), FApp::ShouldUseThreadingForPerformance());
+
+    // Init Sim main components
+    InitSim();
+}
+
+void ARRROS2GameMode::InitSim()
```
yuokamoto, line 38: what is the intension of this method?

Tadinu: Ah right, I thought you meant [`InitGame()`](https://github.com/EpicGames/UnrealEngine/blob/release/Engine/Source/Runtime/Engine/Classes/GameFramework/GameModeBase.h#L61).
This one `InitSim()` as made `virtual` is mostly for child game modes to override and add their possible custom Sim-global component initializing.

yuokamoto (approved):

PR #38, opened 2022-01-17
fix to namespace

larryng-rapyuta:
Linked to https://github.com/rapyuta-robotics/io_amr_UE/pull/57

### DIFF
--- a/Source/RapyutaSimulationPlugins/Private/Tools/ROS2Spawnable.cpp
+++ b/Source/RapyutaSimulationPlugins/Private/Tools/ROS2Spawnable.cpp
@@ -8,7 +8,7 @@
 void UROS2Spawnable::InitializeParameters(const FROSSpawnEntity_Request& InRequest)
 {
     SetName(InRequest.state_name);
-    SetNamespace(InRequest.robot_namespace);
+    SetNamespace(InRequest.robot_namespace.Replace(TEXT("/"),TEXT("")));
 }
 
 void UROS2Spawnable::SetName(const FString& InName)


Source/RapyutaSimulationPlugins/Private/Tools/ROS2Spawnable.cpp
```diff
@@ -8,7 +8,7 @@
 void UROS2Spawnable::InitializeParameters(const FROSSpawnEntity_Request& InRequest)
 {
     SetName(InRequest.state_name);
-    SetNamespace(InRequest.robot_namespace);
+    SetNamespace(InRequest.robot_namespace.Replace(TEXT("/"),TEXT("")));
```
Tadinu, line 11: It looks like we would not want any `/` at all in `robot_namespace`, not just as a leading character. If so, the PR description could have a bit of clarifying update I guess.

Tadinu (approved):

Source/RapyutaSimulationPlugins/Private/Tools/ROS2Spawnable.cpp
Tadinu: It looks like we would not want any `/` at all in `robot_namespace`, not just as a leading character. If so, the PR description could have a bit of clarifying update I guess.
Also, maybe this could use a log message to let users know about the trimming?

yuokamoto: we can create utility function for sanity check for this in the future, I think. 
Because of ros convention, there maybe multiple place to check this.

PR #39, opened 2022-01-29
BaseSensor and BaseSensorPublisher class

yuokamoto:
BaseSensor as parent types for 2D/3DLidar & Camera components
& BaseSensorPublisher as parent types for LaserScan & PointCloud2 publishers.


Tadinu (approved):
Thanks, this is really neat generalization!

Source/RapyutaSimulationPlugins/Private/Sensors/RR2DLidarComponent.cpp
```diff
@@ -6,11 +6,13 @@
 #include "Kismet/KismetMathLibrary.h"
 #include "PhysicalMaterials/PhysicalMaterial.h"
 
+#include "Tools/RRROS2LidarPublisher.h"
+
 #include <limits>
```
Tadinu, line 11: // UE, //rclUE, //RapyutaSimulationPlugins for avoiding auto clang-format reordering?

Tadinu: // UE, //std, //RapyutaSimulationPlugins for avoiding auto clang-format reordering?

PR #42, opened 2022-02-02
Spawn with tag

yuokamoto:
add tags with spawn. will be used to search map origin to floor changes

pr dependency
https://github.com/rapyuta-robotics/rclUE/pull/32
https://github.com/rapyuta-robotics/UE_msgs/pull/5
https://github.com/rapyuta-robotics/UE_msgs/pull/6

### DIFF
--- a/Source/RapyutaSimulationPlugins/Public/Tools/SimulationState.h
+++ b/Source/RapyutaSimulationPlugins/Public/Tools/SimulationState.h
@@ -15,6 +15,18 @@
 
 #include "SimulationState.generated.h"
 
+USTRUCT()
+struct RAPYUTASIMULATIONPLUGINS_API FActorsWithTag
+{
+    GENERATED_BODY()
+
+    UPROPERTY()
+    FName tag;
+
+    UPROPERTY()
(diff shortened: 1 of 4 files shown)


Source/RapyutaSimulationPlugins/Public/Tools/SimulationState.h
```diff
@@ -15,6 +15,18 @@
 
 #include "SimulationState.generated.h"
 
+USTRUCT()
+struct RAPYUTASIMULATIONPLUGINS_API FActorsWithTag
+{
+    GENERATED_BODY()
+
+    UPROPERTY()
+    FName tag;
```
tetorea, line 24: is this tag variable necessary if FActorsWithTag is used in a TMap?
Or are there future case where it won't be the case?

tetorea (approved):

Source/RapyutaSimulationPlugins/Public/Tools/SimulationState.h
yuokamoto: I think, I can remove

PR #50, opened 2022-03-10
[ExternalTest] Fix cmd_vel publishing

Tadinu:
* `cmd_vel` is published through a timer's async callback.
* The current `rclpy's spin_once`, due to missing `timeout_sec` arg specifying, causes itself to block forever if no Twist msg has been published.
Besides, since it also uses the Global executor, which could not be shut down locally, its spinning just goes on without end, causing the Node itself to be not fully destroyed.
& also in order to duly invoke `twist_robot()` as timer's callback, `asyncio.sleep()`, which requires a running event loop, is added.
Thus `SingleThreadedExecutor` is used for calling `spin_once` instead.


Tadinu:
@yuokamoto Should I merge this PR?

yuokamoto (approved):

PR #52, opened 2022-03-22
Add core sources

Tadinu:
Core Sources accomodating synthetic data generation + Gazebo world loader
* `ARRGameMode`, inheriting from `ARRROS2GameMode`, used for game modes utilizing early loaded assets for use in runtime
* `URRGameInstance, ARRGameState, ARRGamePlayerController, URR~Utils, URR~Common` service objects
* `URRSceneInstance, ARRSceneDirector` as base classes for scene-wise operations, which will individualize world mesh loading for each scene, eg warehouse floor.
* `ARRBaseActor` base actor for mesh & robot actors
* `ARRMeshActor`, housing StaticMeshComponent for now and probably also ProceduralMeshComp/RuntimeMeshComp then


Source/RapyutaSimulationPlugins/Public/Core/RRActorCommon.h
```diff
@@ -0,0 +1,523 @@
...
+#define RAPYUTA_SIM_DEBUG (0)
+#define RAPYUTA_SIM_VISUAL_DEBUG (0)
+
+class URRStaticMeshComponent;
+class ARRGameMode;
+class ARRGameState;
+class URRCoreUtils;
+
+USTRUCT()
+struct RAPYUTASIMULATIONPLUGINS_API FRapyutaSimVersionInfo
```
yuokamoto, line 30: Do we need to update hardcoded value here for updating versioning everytime?
We will add Changelog anyway and etc and are there any other option to get version information?

Source/RapyutaSimulationPlugins/Public/Core/RRActorCommon.h
```diff
@@ -0,0 +1,523 @@
...
+    INI,
+    IMAGE_JPG,
+    IMAGE_PNG,
+    IMAGE_EXR,
+
+    // 3D Description Format
+    URDF,
+    SDF,
+    GAZEBO_WORLD,
+    GAZEBO_YAML,
```
yuokamoto: it should be just yaml. 
Are there any gazebo specific yaml formatl.

Source/RapyutaSimulationPlugins/Public/Core/RRActorCommon.h
```diff
@@ -0,0 +1,523 @@
...
+UENUM()
+enum class ERRShapeType : uint8
+{
+    INVALID,
+    PLANE,
+    BOX,
+    CYLINDER,
+    SPHERE,
+    MESH,
+    TOTAL
```
yuokamoto, line 83: maybe better to have CAPSULE which is primitive of UE4?

Source/RapyutaSimulationPlugins/Private/Tools/SimulationState.cpp
```diff
@@ -241,48 +245,55 @@ void ASimulationState::SpawnEntitySrv(UROS2GenericSrv* Service)
...
+            UGameplayStatics::FinishSpawningActor(newEntity, worldTransf);
+            AddEntity(newEntity);
 
-        UE_LOG(LogRapyutaCore, Warning, TEXT("New Spawned Entity Name: %s"), *NewEntity->GetName());
+            UE_LOG(LogRapyutaCore, Warning, TEXT("New Spawned Entity Name: %s"), *newEntity->GetName());
+        }
+        else
+        {
+            Response.bSuccess = false;
+            UE_LOG(LogRapyutaCore, Error, TEXT("Entity spawning failed - [%s] given name actor already exists!"), *entityName);
```
yuokamoto: Can you put same msg to Response.status_message

Source/RapyutaSimulationPlugins/Private/Core/RRROS2GameMode.cpp
```diff
@@ -59,6 +60,6 @@ void ARRROS2GameMode::InitROS2()
     ClockPublisher->InitializeWithROS2(ROS2Node);
 
     // Simulation state
-    SimulationState = currentWorld->SpawnActor<ASimulationState>();
+    SimulationState = currentWorld->SpawnActor<ASimulationState>(SimulationStateClass);
```
yuokamoto, line 63: what is the difference?

Source/RapyutaSimulationPlugins/Public/Core/RRMathUtils.h
```diff
@@ -0,0 +1,25 @@
...
+
+#include "RRMathUtils.generated.h"
+
+UCLASS()
+class RAPYUTASIMULATIONPLUGINS_API URRMathUtils : public UBlueprintFunctionLibrary
+{
+    GENERATED_BODY()
+
+public:
+    FORCEINLINE static FVector ConvertHandedness(const FVector& InLocation)
```
yuokamoto: just from curiousity, 
is handedness correct terminology?
@chalice-graeme

Source/RapyutaSimulationPlugins/Public/Core/RRGameMode.h
```diff
@@ -0,0 +1,68 @@
+// Copyright 2020-2021 Rapyuta Robotics Co., Ltd.
```
yuokamoto, line 1: Do we need RRGameMode? Don't we have RRInternalGamemode?

Source/RapyutaSimulationPlugins/Private/Core/RRGameMode.cpp
```diff
@@ -0,0 +1,214 @@
...
+    }
+    else
+    {
+        UE_LOG(LogTemp, Display, TEXT("PHYSICS RUNS IN A SINGLE THREAD!"));
+    }
+}
+
+void ARRGameMode::ConfigureSimInPlay()
+{
+    if (CaptureResolution.Size() > 0)
```
yuokamoto, line 95: what following settings for?
make it full screen always?

Source/RapyutaSimulationPlugins/Private/Core/RRGameMode.cpp
```diff
@@ -0,0 +1,214 @@
...
+    {
+        UGameUserSettings* gameSettings = GEngine->GetGameUserSettings();
+        gameSettings->SetFullscreenMode(EWindowMode::Windowed);
+        gameSettings->SetScreenResolution(FIntPoint(CaptureResolution.X, CaptureResolution.Y));
+        gameSettings->ApplyResolutionSettings(false);
+    }
+
+    URRCoreUtils::ExecuteConsoleCommand(this, URRCoreUtils::CMD_RHI_ENABLE_GPU_CAPTURE_OPTIONS);
```
yuokamoto: what these consolecommands for?
Do we need to execute from consolecommand?

Source/RapyutaSimulationPlugins/Private/Core/RRPlayerController.cpp
```diff
@@ -0,0 +1,65 @@
...
+{
+    Super::BeginPlay();
+
+    GameMode = URRCoreUtils::GetGameMode<ARRGameMode>(this);
+    check(GameMode);
+
+    GameState = Cast<ARRGameState>(UGameplayStatics::GetGameState(this));
+    check(GameState);
+
+    GameInstance = Cast<URRGameInstance>(UGameplayStatics::GetGameInstance(this));
```
yuokamoto: CoreUtils::GetGameInstance?

Source/RapyutaSimulationPlugins/Private/Core/RRGameInstance.cpp
```diff
@@ -0,0 +1,52 @@
...
+#include "Core/RRGameSingleton.h"
+#include "Core/RRMathUtils.h"
+
+void URRGameInstance::PreloadContentForURL(FURL InURL)
+{
+    UE_LOG(LogRapyutaCore, Display, TEXT("[URRGameInstance]: Preload Content for Map [%s]"), *InURL.Map)
+    Super::PreloadContentForURL(InURL);
+}
+
+AGameModeBase* URRGameInstance::CreateGameModeForURL(FURL InURL, UWorld* InWorld)
```
yuokamoto, line 17: what this for and where this is used?

Source/RapyutaSimulationPlugins/Public/Core/RRActorCommon.h
```diff
@@ -0,0 +1,523 @@
...
+                   TEXT("There are more than 255 CustomDepthStencil values having been assigned!"
+                        "Depth Segmentation Mask info would be duplicated!"));
+        }
+        return ++sLatestCustomDepthStencilValue;
+    }
+};
+
+class ARRSceneDirector;
+class ARRPlayerController;
+UCLASS()
```
yuokamoto, line 273: What this and SceneDirector classes for?

Source/RapyutaSimulationPlugins/Public/Core/RRGameState.h
```diff
@@ -0,0 +1,82 @@
...
+
+// RapyutaSimulationPlugins
+#include "Core/RRActorCommon.h"
+
+#include "RRGameState.generated.h"
+
+class ARRGameMode;
+class URRGameInstance;
+class ARRPlayerController;
+class ARRMeshActor;
```
yuokamoto: gamestate need MeshActor?

Source/RapyutaSimulationPlugins/Public/Core/RRUObjectUtils.h
```diff
@@ -0,0 +1,382 @@
...
+
+// RapyutaSimulationPlugins
+#include "Core/RRActorCommon.h"
+#include "Core/RRBaseActor.h"
+#include "Core/RRCoreUtils.h"
+#include "Core/RRThreadUtils.h"
+
+#include "RRUObjectUtils.generated.h"
+
+class ARRMeshActor;
```
yuokamoto: Need ARRMEshActor?

Source/RapyutaSimulationPlugins/Public/Core/RRThreadUtils.h
```diff
@@ -0,0 +1,44 @@
+// Copyright 2020-2021 Rapyuta Robotics Co., Ltd.
```
yuokamoto, line 1: What is the reason of creating this class?
is this just wrapper?

Source/RapyutaSimulationPlugins/Public/Core/RRActorCommon.h
```diff
@@ -0,0 +1,523 @@
...
+class RAPYUTASIMULATIONPLUGINS_API URRActorCommon : public UObject
+{
+    GENERATED_BODY()
+    // !! [ActorCommon] OBJECT COULD ONLY BE ACCESSED FROM BEGINPLAY(),
+    // THUS SPECIFC SIM START-UP RELATED CONTENTS & HANDLINGS, WHICH HAPPENS BEFORE BEGINPLAY(), SHOULD BE MOVED TO GameMode or
+    // GameInstance.
+    friend class URRCoreUtils;
+
+private:
+    static TMap<int8, URRActorCommon*> SActorCommonList;
```
yuokamoto, line 188: start with captial s is ue coding standards?

Source/RapyutaSimulationPlugins/Public/Core/RRCoreUtils.h
```diff
@@ -0,0 +1,381 @@
...
+
+        if (!bResult && bIsLogged)
+        {
+            UE_LOG(LogTemp, Error, TEXT("Command line argument not found under the name [%s]"), InArgName);
+        }
+        return bResult;
+    }
+
+    template<typename T>
+    static FORCEINLINE constexpr bool ParseParams(const FString& InParams,
```
yuokamoto: where this function used?
if this is or  commandline execution,   it maybe better to rename with commandline~, since ParseParam is too generic name

Source/RapyutaSimulationPlugins/Private/Core/RRUObjectUtils.cpp
```diff
@@ -0,0 +1,144 @@
...
+#include "Components/ActorComponent.h"
+#include "GameFramework/Actor.h"
+
+// RapyutaSimulationPlugins
+#include "Core/RRBaseActor.h"
+#include "Core/RRMathUtils.h"
+
+void URRUObjectUtils::SetupActorTick(AActor* InActor, bool bIsTickEnabled, float InTickInterval)
+{
+    // Tick if not required could be disabled to improve performance
```
yuokamoto, line 13: Just from curiosity, how much this can improve performance?
cc @chalice-graeme

Source/RapyutaSimulationPlugins/Private/Core/RRUObjectUtils.cpp
```diff
@@ -0,0 +1,144 @@
...
+}
+
+void URRUObjectUtils::SetupComponentTick(UActorComponent* InComponent, bool bIsTickEnabled)
+{
+    InComponent->PrimaryComponentTick.bCanEverTick = bIsTickEnabled;
+    InComponent->PrimaryComponentTick.bStartWithTickEnabled = bIsTickEnabled;
+    InComponent->bTickInEditor = bIsTickEnabled;
+}
+
+void URRUObjectUtils::SetupDefaultRootComponent(AActor* InActor)
```
yuokamoto, line 45: Just from curiosity, do you know why UE has two diifferent methods for functions inside/outside of constructors?
cc @chalice-graeme

Source/RapyutaSimulationPlugins/Private/Tools/SimulationState.cpp
```diff
@@ -241,48 +245,55 @@ void ASimulationState::SpawnEntitySrv(UROS2GenericSrv* Service)
 {
```
yuokamoto, line 245: Is it possible to integrate witth URRUObjectUtils::SpawnSimActor?(not necessary but curious)

Source/RapyutaSimulationPlugins/Public/Core/RRBaseActor.h
```diff
@@ -0,0 +1,75 @@
...
+
+class ARRGameMode;
+class ARRGameState;
+class ARRPlayerController;
+
+/* Base actor class for all Rapyuta Sim actors:
+ * - Be assigned a sim mode instance id
+ * - Provide accessible handles to the common Game framework objects (GameInstance, GameMode, GameState) and
+ * the corresponding [ActorCommon] of the Sim Mode instance it belongs to.
+ * - Does not necessarily has a USceneComponent as Root!
```
yuokamoto: can you write down the merit comes from not having a USceneCompoennt.

yuokamoto (commented):
Any policy for differentiate between CoreUtils, GeneralUtils, ObjectUtils?

Source/RapyutaSimulationPlugins/Public/Core/RRActorCommon.h
Tadinu: Yeap, so Sim version is like the signature of the sources thus should be an inherent part of the codes. But typically the Version is useful to help identify the output data version and thus mostly meaningful upon synthetic data generation-related changes, like to the data structure, are added.

Source/RapyutaSimulationPlugins/Private/Core/RRROS2GameMode.cpp
Tadinu: The spawned actor will be of the type indicated by `SimulationStateClass` (similarly as in `NewObject`).
This would facilitate the creation of instances of an `ASimulationState`'s child class like `ARRInternalSimulationState`, which is required for the custom handling of the internal service like `SpawnWorld`.

Source/RapyutaSimulationPlugins/Public/Core/RRGameMode.h
Tadinu: * `RRGameMode` is mostly to trigger early resource loading, which is commonly used for both `RapyutaRobotImporter /RatapyutaSimInternal(GazeboWorldLoader)`.
* `RRInternalGameMode` is to configure `GameStateClass, PlayerControllerClass, SimulationStateClass`, allowing custom handling for internal-specific resources & functionalities used by `RapyutaSimInternal` repo only.

Source/RapyutaSimulationPlugins/Private/Core/RRGameMode.cpp
Tadinu: It is mostly for scene capturing size as in data generation only, which will be configured by users. Using it also for the screen resolution is only to give users a sense of consistency between the Sim main screen size vs output image size.

Source/RapyutaSimulationPlugins/Private/Core/RRGameMode.cpp
Tadinu: Yeah, so that is the UE built-in console command, not the OS provided console, typically used to quickly configure graphics options. This like above is part of synth data generation.

Source/RapyutaSimulationPlugins/Private/Core/RRGameInstance.cpp
Tadinu: This is a UE-provided API, used somewhere, in the process of Sim initialization. In some cases, we really need to know about the details of the order of game framework actors/services instantiation, thus logging them out.

Source/RapyutaSimulationPlugins/Public/Core/RRActorCommon.h
Tadinu: As in the PR description, this is the base class managing scene-wise operations, which will individualize world mesh loading for each scene, eg warehouse floor.
We could imagine that generically Sim could house multiple separate scenes, each instance of which has:
* `Scene director` for defining actor spawning, manipulating plan for specific scene settings/arrangement (typically customized in a synthetic data generation scene). In case of world loader, it could possibly act as the manager to define how different world instances are spawned into the scene also.
* `~Common` object handling shared operations done (like querying, housing common resources/entities, etc) in the scene.
* `~PlayerController` for camera operations in that scene only.
The idea is to maximize the throughput of a synthetic data generation app, by letting multiple scene instances run in parallel to optimize away inherent down-time in each scence's operation (like waiting for an object to settle on ground or the scene capturing itself).

Source/RapyutaSimulationPlugins/Public/Core/RRThreadUtils.h
Tadinu: Yeah, let me remove wrapper ones, I added it in early days just out of not always remembering the UE built in class name.

Source/RapyutaSimulationPlugins/Public/Core/RRActorCommon.h
Tadinu: Yeah, `S` actually stands for static, which is not fully recommended thus needs full caution on handling, and capital is for a UE member.

Source/RapyutaSimulationPlugins/Public/Core/RRCoreUtils.h
Tadinu: Right, it is not used yet, and will be used for running a custom automation test commandlet, or typically when users want to pass custom parameter to our Sim. Yeap, `ParseCommandlineParams` sounds better.

Source/RapyutaSimulationPlugins/Private/Core/RRUObjectUtils.cpp
Tadinu: Ticks themselves are calls invoked frequently like ROS topic publication, thus generally, a just-enough policy should be encouraged in that sense I suppose.

Source/RapyutaSimulationPlugins/Private/Tools/SimulationState.cpp
Tadinu: Probably yes, just there is one point I am still not clear about the use of `SpawnActorDeferred` here, which makes this spawning a bit too specific, thus not as a straight spawn as `URRUObjectUtils::SpawnSimActor()`.

Source/RapyutaSimulationPlugins/Public/Core/RRBaseActor.h
Tadinu: Sure, SceneComponent deals with object poses and their accompanying specific transform calculations. Also, in the hierarchy, `USceneComponent` is a child class of `UActorComponent`, so factually the parent classes are less expensive than child classes instances. 
Since this is the base actor class for all in the Sim, it could parent both mesh-based and non-meshed based ones (like the SimulationState), it should be as lean as possible.

Tadinu: Sure, SceneComponent deals with object poses and their accompanying specific transform calculations. Also, in the hierarchy, `USceneComponent` is a child class of `UActorComponent`, so factually the parent classes are less expensive than child ones. 
Since this is the base actor class for all in the Sim, it could parent both mesh-based and non-meshed based ones (like the SimulationState), it should be as lean as possible.

Source/RapyutaSimulationPlugins/Private/Core/RRGameMode.cpp
yuokamoto: Can this crash the UE4 with machine without GPU?
(if it do not cause crash in jenkins, it should be ok)

Tadinu: Right, in terms of API friendliness, UE should not trap users into crash by some mis-configured value and issue use a Warning log instead. Generally, since UE has numerous configs related to GPU (like for ray-tracing), it should properly handle those config values by RHI as in line with the GPU status.
But yeah for sure, let me move this to `RRInternalGameMode` since it is part of scene capturing. Btw, the handling callback for this configure is [here](https://github.com/EpicGames/UnrealEngine/blob/release/Engine/Source/Runtime/RHI/Private/DynamicRHI.cpp#L380).

Source/RapyutaSimulationPlugins/Public/Core/RRMathUtils.h
Tadinu: https://en.wikipedia.org/wiki/Handedness & in the context of GL, it is about whether right-hand or left-hand is used to represent the coordinate axes.

Source/RapyutaSimulationPlugins/Private/Core/RRUObjectUtils.cpp
Tadinu: One of the most peculiar traits of UE that I am not a fan of (and thus do not favor BP dev at all), but I think it might be to support child BP class instantiation, which requires member objects to be created as default objects as inside Ctor (so they could be configured/modified in the BP Editor) vs BeginPlay, where only NewObject is used.

Tadinu:
> Any policy for differentiate between CoreUtils, GeneralUtils, ObjectUtils?

Right, so the idea is to encourage the categorization of Utils for sake of management and handiness in use
* CoreUtils: UE Core & std native type related services (string manipulation, IO, Timer, Image Wrapper, etc.)
* UObjectUtils: Any UObject-centric services (like Actor/Component creation, querying, configuring, etc)
* GeneralUtils: for other not-yet-decided at the time, but anw should be put to specialized ~Utils class later if possible

Source/RapyutaSimulationPlugins/Public/Core/RRMathUtils.h
yuokamoto: This method may not necessary since we have ConversionUtils

Tadinu: Well, it is actually needed and used currently to convert coordinate in Assimp to UE. ConversionUtils only do ROS<->UE conversion and implicitly include this handedness, but not all cases that we need to do 100 cm<->m conversion.
This function may not rightfully belong to MathUtils and rather be moved to ConversionUtils, but I prefer to keep the class since anyway upcoming methods will come soon.

yuokamoto (approved):

PR #53, opened 2022-03-23
Server Client setup

larryng-rapyuta:


yuokamoto:
seems good. 
it maybe better to create playercontroller base class which can be used for robot arm and amr in the future. odom publisher and etc is only used for mobile robot.

Tadinu:
Speaking of which, I happened to add `ARRPlayerControler` class [here](https://github.com/rapyuta-robotics/RapyutaSimulationPlugins/blob/AddCore/Source/RapyutaSimulationPlugins/Public/Core/RRPlayerController.h) in my PR, so it could be reused for the purpose I guess.

Source/RapyutaSimulationPlugins/Public/Tools/SimulationStateData.h
```diff
@@ -0,0 +1,65 @@
...
+        GENERATED_BODY()
+
+        UPROPERTY()
+        TArray<AActor*> Actors;
+};
+
+class UROS2GenericSrv;
+
+UCLASS()
+class RAPYUTASIMULATIONPLUGINS_API ASimulationStateData : public AGameState
```
Tadinu, line 24: It looks like this class could inherit directly from `UObject`?

Source/RapyutaSimulationPlugins/Private/Tools/SimulationState.cpp
```diff
@@ -83,26 +84,26 @@ void ASimulationState::GetEntityStateSrv(UROS2GenericSrv* Service)
 
     FROSGetEntityState_Response Response;
     Response.success = false;
-    if (Entities.Contains(Request.name))
+    if (SimulationStateData->Entities.Contains(Request.name))
```
chalice-graeme: Worth pointing out that there are _a lot_ of map lookups for this entity and Request.reference_frame. It would be better to pull the actor out into a local variable and do operations on it that way.

Source/RapyutaSimulationPlugins/Private/RRROS2GameMode.cpp
```diff
@@ -60,5 +80,9 @@ void ARRROS2GameMode::InitROS2()
 
     // Simulation state
     SimulationState = currentWorld->SpawnActor<ASimulationState>();
-    SimulationState->Init(ROS2Node);
+//    ASimulationStateData* SimulationStateData = GetWorld() != NULL ? GetWorld()->GetGameState<ASimulationStateData>() : NULL;
+    SimulationStateData = currentWorld->SpawnActor<ASimulationStateData>();
```
chalice-graeme, line 83: Does this actually need to be an actor? It's quite heavyweight if it's just a data storage container. If it needs to be serialized by Unreal a UObject derived class would be better, I think.

PR #55, opened 2022-04-25
Add procedural mesh comp

Tadinu:
[Procedural mesh](https://github.com/SiggiG/ProceduralMeshDemos) allows Sim to build mesh actor on the fly in runtime, meaning its meshes could be created directly external 3D cads, without using static mesh and thus unburden Sim of packaging static mesh + accompanying textures.

This PR thus make mesh actor to be composed of Procedural mesh components by default, though Static mesh comps are still supported. Specifically,
* `URRProceduralMeshComponent` inheriting `UProceduralMeshComponent`, async mesh loading + async collision cooking by default
* Add `ERRResourceDataType::UE_BODY_SETUP` for dynamically cooked collision `UBodySetup` object, stored in `URRGameSingleton::ResourceStore` for re-using in runtime.
* `FRRMeshData + URRMeshUtils` for housing mesh data loaded from 3D cad, collectively stored in `FRRMeshData::MeshDataStore` for same reason.

Regarding the mesh loading api, we use [Assimp](https://github.com/assimp/assimp)([BSD license](http://www.assimp.org/index.php?id=11:faq)), of which release header files + lib are added directly as 3rd dependency.
This is though not ideal but adopted for reasons:
* To help CI avoid pulling Assimp sources (around ~150MB), on every build
* Also, as observed, Assimp, only working as a static lib, must be built beforehand, which takes certain amount of time
* To make life a bit easier for client repos' users

**Notes:**
As using async cooking, there is an error log related to `PhysX` spewed out at Sim shutdown:
```
LogPhysicsCore: Error: PHYSX: (D:/Build/++UE4/Sync/Engine/Source/ThirdParty/PhysX3/PxShared/src/foundation/src/PsFoundation.cpp 180) eINVALID_OPERATION : Foundation destruction failed due to pending module references. Close/release all depending modules first.
```
* The reason is unclear yet, but as observed, it does not impact created proc-mesh actor. This probably comes from lingering cooked collision data that is not cleanly deallocated.
* In our Sim, a specific measure has been added to make sure ProceMeshComp's BodySetup collision info is cleared.
Refer to `FRRResourceInfo::Finalize()`, which is invoked in `URRGameSingleton::FinalizeResources()` by `ARRGameState::EndPlay()`

### DIFF
--- a/Source/RapyutaSimulationPlugins/Private/Core/RRActorCommon.cpp
+++ b/Source/RapyutaSimulationPlugins/Private/Core/RRActorCommon.cpp
@@ -66,17 +66,26 @@ bool URRSceneInstance::IsValid(bool bIsLogged) const
 // ===================================================================================================================================
 // [FRRActorSpawnInfo] --
 //
-FRRActorSpawnInfo::FRRActorSpawnInfo(const FString& InUniqueName,
+FRRActorSpawnInfo::FRRActorSpawnInfo(const FString& InEntityModelName,
+                                     const FString& InUniqueName,
                                      const FTransform& InTransform,
                                      const TArray<FString>& InMeshMaterialNameList,
                                      bool bIsStationary,
                                      bool bIsPhysicsEnabled,
                                      bool bIsCollisionEnabled)
 {
-    (*this)(InUniqueName, InTransform, EMPTY_STR, InMeshMaterialNameList, bIsStationary, bIsPhysicsEnabled, bIsCollisionEnabled);
+    (*this)(InEntityModelName,
+            InUniqueName,
+            InTransform,
+            EMPTY_STR,
+            InMeshMaterialNameList,
+            bIsStationary,
+            bIsPhysicsEnabled,
+            bIsCollisionEnabled);
 }
 
-FRRActorSpawnInfo::FRRActorSpawnInfo(const FString& InUniqueName,
+FRRActorSpawnInfo::FRRActorSpawnInfo(const FString& InEntityModelName,
+                                     const FString& InUniqueName,
                                      const FTransform& InTransform,
                                      const FString& InMeshUniqueName,
                                      const TArray<FString>& InMeshMaterialNameList,
@@ -84,18 +93,26 @@ FRRActorSpawnInfo::FRRActorSpawnInfo(const FString& InUniqueName,
                                      bool bIsPhysicsEnabled,
                                      bool bIsCollisionEnabled)
 {
-    (*this)(
-        InUniqueName, InTransform, InMeshUniqueName, InMeshMaterialNameList, bIsStationary, bIsPhysicsEnabled, bIsCollisionEnabled);
+    (*this)(InEntityModelName,
+            InUniqueName,
+            InTransform,
+            InMeshUniqueName,
+            InMeshMaterialNameList,
+            bIsStationary,
+            bIsPhysicsEnabled,
+            bIsCollisionEnabled);
 }
 
-FRRActorSpawnInfo::FRRActorSpawnInfo(const FString& InUniqueName,
+FRRActorSpawnInfo::FRRActorSpawnInfo(const FString& InEntityModelName,
+                                     const FString& InUniqueName,
                                      const FTransform& InTransform,
                                      const TArray<FString>& InMeshUniqueNameList,
                                      const TArray<FString>& InMeshMaterialNameList,
                                      bool bIsStationary,
                                      bool bIsPhysicsEnabled,
                                      bool bIsCollisionEnabled)
-    : UniqueName(InUniqueName),
+    : EntityModelName(InEntityModelName),
+      UniqueName(InUniqueName),
       Transform(InTransform),
       MeshUniqueNameList(InMeshUniqueNameList),
       MeshMaterialNameList(InMeshMaterialNameList),
@@ -105,7 +122,8 @@ FRRActorSpawnInfo::FRRActorSpawnInfo(const FString& InUniqueName,
 {
 }
 
-void FRRActorSpawnInfo::operator()(const FString& InUniqueName,
+void FRRActorSpawnInfo::operator()(const FString& InEntityModelName,
+                                   const FString& InUniqueName,
                                    const FTransform& InTransform,
                                    const FString& InMeshUniqueName,
                                    const TArray<FString>& InMeshMaterialNameList,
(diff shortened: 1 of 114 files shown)


yuokamoto:
> Regarding the mesh loading api, we use [Assimp](https://github.com/assimp/assimp)([BSD license](http://www.assimp.org/index.php?id=11:faq)), of which release header files + lib are added directly as 3rd dependency. This is though not ideal but adopted for reasons:
> 
> * To help CI avoid pulling Assimp sources (around ~150MB), on every build
> * Also, as observed, Assimp, only working as a static lib, must be built beforehand, which takes certain amount of time
> * To make life a bit easier for client repos' users

@chalice-graeme is this same issue as ROS2 libraries?

Source/RapyutaSimulationPlugins/Private/Core/RRActorCommon.cpp
```diff
@@ -66,36 +66,53 @@ bool URRSceneInstance::IsValid(bool bIsLogged) const
...
+FRRActorSpawnInfo::FRRActorSpawnInfo(const FString& InEntityModelName,
+                                     const FString& InUniqueName,
                                      const FTransform& InTransform,
                                      const TArray<FString>& InMeshMaterialNameList,
                                      bool bIsStationary,
                                      bool bIsPhysicsEnabled,
                                      bool bIsCollisionEnabled)
 {
-    (*this)(InUniqueName, InTransform, EMPTY_STR, InMeshMaterialNameList, bIsStationary, bIsPhysicsEnabled, bIsCollisionEnabled);
+    (*this)(InEntityModelName,
```
yuokamoto: what is the difference from delegating constructor?
https://docs.microsoft.com/en-us/cpp/cpp/delegating-constructors?view=msvc-170

Source/RapyutaSimulationPlugins/Public/Core/RRCoreUtils.h
```diff
@@ -354,6 +354,30 @@ class RAPYUTASIMULATIONPLUGINS_API URRCoreUtils : public UBlueprintFunctionLibra
         return InWorld->GetTimerManager().SetTimerForNextTick(MoveTemp(InCallback));
     }
 
+    template<typename T>
```
yuokamoto, line 361: where those utils are used?

Source/RapyutaSimulationPlugins/Private/Core/RRActorCommon.cpp
Tadinu: Yeah so this so-called [functor(or function object)](https://docs.microsoft.com/en-us/cpp/standard-library/function-objects-in-the-stl) is used to handily invoke an operation on an object through its `operator()`, which looks like a function and thus comes that name.
Meanwhile, delegating ctor can only be used in the ctor's initializer list, and the [catch](https://stackoverflow.com/a/12190126) is that no other members could be added, meaning they have to be initialized by assignment in ctor.

Source/RapyutaSimulationPlugins/Public/Core/RRCoreUtils.h
Tadinu: This is used specifically in [URRProceduralMeshComponent](https://github.com/rapyuta-robotics/RapyutaSimulationPlugins/blob/AddProcMeshComp/Source/RapyutaSimulationPlugins/Private/Core/RRProceduralMeshComponent.cpp#L138)  when it needs to wait for async collision cooking to finish in a polling style.

yuokamoto (approved):

Source/RapyutaSimulationPlugins/Private/Core/RRStaticMeshComponent.cpp
```diff
@@ -56,22 +58,23 @@ void URRStaticMeshComponent::Initialize(bool bInIsStationary, bool bInIsPhysicsE
     SetCustomDepthStencilValue(URRActorCommon::GenerateUniqueDepthStencilValue());
 }
 
-void URRStaticMeshComponent::SetMesh(UStaticMesh* Mesh)
+void URRStaticMeshComponent::SetMesh(UStaticMesh* InStaticMesh)
```
chalice-graeme, line 61: Can this be const?

Source/RapyutaSimulationPlugins/Private/Core/RRUObjectUtils.cpp
```diff
@@ -142,3 +144,13 @@ ARRBaseActor* URRUObjectUtils::SpawnSimActor(UWorld* InWorld,
 
     return newActor;
 }
+
+
+UMaterialInstanceDynamic* URRUObjectUtils::CreateMeshCompMaterialInstance(UMeshComponent* InMeshComp,
```
chalice-graeme, line 148: Might want to assert that InMeshComp is not null.

Source/RapyutaSimulationPlugins/Public/Core/RRCoreUtils.h
chalice-graeme: If these are used in an async operation, it would be best to check if the InObj is valid. It's possible there is a circumstance where it is deleted during a pending operation. At very least a debug-only check and log would be good.

Source/RapyutaSimulationPlugins/RapyutaSimulationPlugins.Build.cs
```diff
@@ -1,19 +1,66 @@
...
+
+    private string GetLibPath(string InLibName)
+    {
+        return Path.Combine(ThirdPartyPath, InLibName);
+    }
+
+    private void AddLib(string InLibPath, string InLibSymLinkName, string InLibName = "")
+    {
+        string libFullPath = Path.Combine(InLibPath, "release/lib", InLibSymLinkName);
+        if (InLibSymLinkName.EndsWith(".a") || InLibSymLinkName.EndsWith(".so"))
```
chalice-graeme: Just a note, that this check is insufficient for non-Unix-like platforms. On Windows these files would end in .obj and .dll instead. Might be worth adding a comment about portability.

Source/RapyutaSimulationPlugins/Private/Core/RRStaticMeshComponent.cpp
Tadinu: Yeah, I also want to, but `UStaticMeshComponent::SetStaticMesh()` (as well as lots of UE apis) does not have it as passing `UObject*`, which makes itself unrobust, guess only to save the cast.

Tadinu: Yeah, I also want to, but `UStaticMeshComponent::SetStaticMesh()` (as well as lots of UE apis) does not have it as receiving `UObject*`, which makes itself unrobust, guess only to save the cast.

tetorea (approved):

PR #58, opened 2022-05-21
add missing InitMovementComponent to BeginPlay of RobotVehicleMovemen…

yuokamoto:
…tComponent

### DIFF
--- a/Source/RapyutaSimulationPlugins/Private/Drives/RobotVehicleMovementComponent.cpp
+++ b/Source/RapyutaSimulationPlugins/Private/Drives/RobotVehicleMovementComponent.cpp
@@ -9,6 +9,12 @@
 #include "GameFramework/PlayerController.h"
 #include "Kismet/KismetMathLibrary.h"
 
+void URobotVehicleMovementComponent::BeginPlay()
+{
+    Super::BeginPlay();
+    InitMovementComponent();
+}
+
 void URobotVehicleMovementComponent::Initialize()
 {
     GaussianRNGPosition = std::normal_distribution<>{NoiseMeanPos, NoiseVariancePos};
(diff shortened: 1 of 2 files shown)


tetorea (approved):

Source/RapyutaSimulationPlugins/Private/Drives/RobotVehicleMovementComponent.cpp
```diff
@@ -9,6 +9,12 @@
 #include "GameFramework/PlayerController.h"
 #include "Kismet/KismetMathLibrary.h"
 
+void URobotVehicleMovementComponent::BeginPlay()
+{
+    Super::BeginPlay();
+    InitMovementComponent();
+}
+
```
Tadinu, line 17: @yuokamoto @tetorea Since `BeginPlay()` has the inherent issue of unknown running order with Level's BP (like we encountered at some point earlier), I wonder if we could call `InitMovementComponent()` in `Initialize()`, which is already called by `ARobotVehicle::InitMoveComponent() from its Super::PostInitializeComponents()`?

tetorea: @Tadinu I think it may not be a problem in this case because what is initialized here is the current state of the robot in the environment. As long as it is called before `UpdateMovement()` then it is fine

yuokamoto: I tired it first but it crashed when I placed robot manually from editor. I guess some initialization process crash if it happen before playing

Tadinu (approved):

PR #62, opened 2022-06-06
Add data synth utils

Tadinu:
io_amr_UE's `DataSynth` branch
By `git submodules status`, it would give:
* RapyutaRobotImporter `DATASYNTHTEXTURE`
* RapyutaSimulationPlugins `AddDataSynthUtils`
* RapyutaSimInternal `AddDataSynth`
* rclUE `devel`

Details
- Add DataSynth Utils
```
RRCamera, RRProceduralMeshComponent
Add RAPYUTA_DATA_SYNTH_USE_ENTITY_STATIC_MESH, RAPYUTA_DATA_SYNTH_ENTITY_USE_PROCEDURAL_MESH
RRGameSingleton MATERIAL_NAME_ASSET_MASTER, MATERIAL_NAME_PROP_MASTER
```
- Runtime StaticMesh using GeometryProcessing plugin, texturing not yet working
- namespace ConversionUtils -> class URRConversionUtils
- Add external texture image loading & coating for procedural mesh actors
- Add utils for metadata (mesh actor bb vertices & 2d projection onto screen)
- Add auto sim shutdown upon data generation finish


yuokamoto:
What is the purpose of RRCamera? Could you add doxygen documentation?

Tadinu:
Yeah, so in order to generate synthetic data, we need:
* UCameraComponent, acting a window to user view of scene, of which the contents are rendered then captured into images.
* An actor for that camera comp to be attached to, which is `ARRCamera`. It is a UE-pure actor as different from the existing `ARRROS2CameraActor`, which is composed of `URRROS2CameraComponent` and has a ros2 node.

Tadinu:
Yeah, so in order to generate synthetic data, we need:
* UCameraComponent, acting as a window to user view of scene, of which the contents are rendered then captured into images.
* An actor for that camera comp to be attached to, which is `ARRCamera`. It is a UE-pure actor as different from the existing `ARRROS2CameraActor`, which is composed of `URRROS2CameraComponent` and has a ros2 node.

Source/RapyutaSimulationPlugins/Public/Core/RRConversionUtils.h
```diff
@@ -13,210 +13,179 @@
...
-                                 const double& InputZ,
-                                 double& OutputX,
-                                 double& OutputY,
-                                 double& OutputZ)
-{
-    OutputX = InputX * 100.0f;
-    OutputY = -InputY * 100.0f;
-    OutputZ = InputZ * 100.0f;
-}
+    FORCEINLINE static void VectorUEToROS(const double& InputX,
```
yuokamoto, line 42: shouldn't add  FORCEINLINE for other methods as well?

Source/RapyutaSimulationPlugins/Public/Core/RRConversionUtils.h
```diff
@@ -13,210 +13,179 @@
...
+    }
+    FORCEINLINE static void VectorROSToUE(const double& InputX,
+                                          const double& InputY,
+                                          const double& InputZ,
+                                          double& OutputX,
+                                          double& OutputY,
+                                          double& OutputZ)
     {
-        return FVector(InLocation.X, -InLocation.Y, InLocation.Z);
+        OutputX = InputX * 100.0f;
```
yuokamoto: 100.f?

Source/RapyutaSimulationPlugins/Public/Core/RRTypeUtils.h
```diff
@@ -17,9 +17,27 @@
 #include "RRTypeUtils.generated.h"
 
 /**
- * @brief UE type related utils
+ * @brief UE/std type related utils
  * 
  */
+template<typename T>
```
yuokamoto, line 23: whrere those struct are used?

Source/RapyutaSimulationPlugins/Public/Core/RRUObjectUtils.h
```diff
@@ -40,13 +42,15 @@ class RAPYUTASIMULATIONPLUGINS_API URRUObjectUtils : public UBlueprintFunctionLi
...
-     * @param InObjectClass 
-     * @param InObjectUniqueName 
+     * @param InOuter
+     * @param InObjectClass
+     * @param InObjectUniqueName
      * @return static UObject*
      *
-     * @sa [CreateDefaultSubobject](https://docs.unrealengine.com/4.27/en-US/API/Runtime/CoreUObject/UObject/UObject/CreateDefaultSubobject/2/)
-     * @sa [NewObject](https://docs.unrealengine.com/4.27/en-US/ProgrammingAndScripting/ProgrammingWithCPP/UnrealArchitecture/Objects/Creation/)
+     * @sa
```
yuokamoto: shouldn't @sa and later in the same line?

Source/RapyutaSimulationPlugins/Public/Core/RRUObjectUtils.h
```diff
@@ -65,11 +69,11 @@ class RAPYUTASIMULATIONPLUGINS_API URRUObjectUtils : public UBlueprintFunctionLi
 
     /**
      * @brief Return true if InParentObj has subobjects with InChildName.
-     * 
-     * @tparam T 
-     * @param InParentObj 
-     * @param InChildName 
-     * @return static bool 
+     *
+     * @tparam T
```
yuokamoto: please remove t at beginning

Source/RapyutaSimulationPlugins/Public/Core/RRUObjectUtils.h
```diff
@@ -169,14 +175,14 @@ class RAPYUTASIMULATIONPLUGINS_API URRUObjectUtils : public UBlueprintFunctionLi
...
-     * 
-     * @tparam T 
-     * @param InActor 
-     * @param InUniqueName 
-     * @param bIsPhysicsEnabled 
-     * @param bIsCollisionEnabled 
-     * @param bIsOverlapEventEnabled 
-     * @return T* 
+     *
+     * @tparam T
```
yuokamoto: same as above

Source/RapyutaSimulationPlugins/Public/Core/RRUObjectUtils.h
```diff
@@ -151,14 +155,16 @@ class RAPYUTASIMULATIONPLUGINS_API URRUObjectUtils : public UBlueprintFunctionLi
...
-     * @param InSocketName 
+     * @param InChildComp
+     * @param InParentComp
+     * @param InRelativeTransf
+     * @param InAttachmentRules
+     * @param InSocketName
      *
-     * @sa [SetupAttachment](https://docs.unrealengine.com/4.27/en-US/API/Runtime/Engine/Components/USceneComponent/SetupAttachment/)
-     * @sa [AttachToComponent](https://docs.unrealengine.com/4.27/en-US/API/Runtime/Engine/Components/USceneComponent/AttachToComponent/)
+     * @sa
```
yuokamoto, line 164: same as above.

Source/RapyutaSimulationPlugins/Public/Core/RRUObjectUtils.h
```diff
@@ -328,21 +324,21 @@ class RAPYUTASIMULATIONPLUGINS_API URRUObjectUtils : public UBlueprintFunctionLi
...
      * @brief Find actor by name. GetAllActors() is expensive.
-     * 
-     * @tparam T 
-     * @param InWorld 
-     * @param InName 
-     * @param InCaseType 
-     * @return T* 
      *
-     * @sa 
+     * @tparam T
```
yuokamoto: same a above

Source/RapyutaSimulationPlugins/Public/Core/RRUObjectUtils.h
```diff
@@ -513,8 +521,46 @@ class RAPYUTASIMULATIONPLUGINS_API URRUObjectUtils : public UBlueprintFunctionLi
...
+    static FVector GetActorExtent(AActor* InActor, bool bOnlyCollidingComponents = true, bool bIncludeFromChildActors = false)
+    {
+        FVector actorOrigin, actorExtent;
+        InActor->GetActorBounds(bOnlyCollidingComponents, actorOrigin, actorExtent, bIncludeFromChildActors);
+        return actorExtent;
+    }
+
+    static FVector GetActorSize(AActor* InActor, bool bOnlyCollidingComponents = true, bool bIncludeFromChildActors = false)
+    {
+        return 2.f * GetActorExtent(InActor, bOnlyCollidingComponents, bIncludeFromChildActors);
```
yuokamoto, line 547: what is 2.f?

yuokamoto (commented):
- Please check doxygen
- Please add missing documentation in this or following pr
- Conirmed it working with io_amr

Source/RapyutaSimulationPlugins/Public/Core/RRTypeUtils.h
Tadinu: It is used [here](https://github.com/rapyuta-robotics/RapyutaSimulationInternal/blob/AddDataSynth/Source/RapyutaSimInternal/Public/SceneCapture/RRCaptureCommon.h#L318) in `URRCaptureCommon::LogJsonProp`

Source/RapyutaSimulationPlugins/Public/Core/RRUObjectUtils.h
Tadinu: It seems clang-format auto break the long line

Source/RapyutaSimulationPlugins/Public/Core/RRUObjectUtils.h
Tadinu: Is that meant for template type, like in this UE [class](https://github.com/EpicGames/UnrealEngine/blob/release/Engine/Plugins/Animation/RigLogic/Source/RigLogicLib/Public/dna/types/ArrayView.h#L46)?

Source/RapyutaSimulationPlugins/Public/Core/RRUObjectUtils.h
Tadinu: So Size = 2.f * Extent(radius)

Source/RapyutaSimulationPlugins/Public/Core/RRConversionUtils.h
Tadinu: They are `BlueprintCallable` so serialized, thus not able to inline.

Source/RapyutaSimulationPlugins/Public/Core/RRUObjectUtils.h
yuokamoto: Can you check it works or not and if it not works, can you avoid change line?

yuokamoto: ah I see.

Source/RapyutaSimulationPlugins/Public/Core/RRCamera.h
```diff
@@ -0,0 +1,84 @@
...
+        verify(FMath::IsWithinInclusive(DistanceRangeInCm.X, 0.f, DistanceRangeInCm.Y));
+        UE_LOG(LogRapyutaCore, Display, TEXT("- HeightRangeInCm: %s"), *HeightRangeInCm.ToString());
+        verify(FMath::IsWithinInclusive(HeightRangeInCm.X, 0.f, HeightRangeInCm.Y));
+        UE_LOG(LogRapyutaCore, Display, TEXT("- HFoVRangeInDegree: %s"), *HFoVRangeInDegree.ToString());
+        verify(FMath::IsWithinInclusive(HFoVRangeInDegree.X, 0.f, HFoVRangeInDegree.Y));
+    }
+};
+
+/**
+ * @brief Standalone camera actor which can be placed in the level with #UCameraComponent.
```
yuokamoto, line 49: Could you add more explanation as you written in pr discussion.

yuokamoto (approved):
Please check @sa works with line change.

Source/RapyutaSimulationPlugins/Private/Core/RRCoreUtils.cpp
```diff
@@ -19,20 +20,36 @@
...
-    {ERRFileType::IMAGE_HDR, TEXT(".hdr")},
-    {ERRFileType::URDF, TEXT(".urdf")},
-    {ERRFileType::SDF, TEXT(".sdf")},
-    {ERRFileType::GAZEBO_WORLD, TEXT(".world")},
-    {ERRFileType::MJCF, TEXT(".mjcf")},
-};
+IImageWrapperModule* URRCoreUtils::SImageWrapperModule = nullptr;
+TMap<ERRFileType, TSharedPtr<IImageWrapper>> URRCoreUtils::SImageWrappers;
+
+const TMap<ERRFileType, const TCHAR*> URRCoreUtils::SimFileExts = {{ERRFileType::NONE, EMPTY_STR},
```
chalice-graeme: This can be a C-style array since the key is an enum.

Source/RapyutaSimulationPlugins/Private/Core/RRCoreUtils.cpp
```diff
@@ -102,14 +130,70 @@ FVector URRCoreUtils::GetSceneInstanceLocation(int8 InSceneInstanceId)
...
+        URRCoreUtils::ExecuteConsoleCommand(URRActorCommon::GetActorCommon(), URRCoreUtils::CMD_MEMORY_REPORT_FULL);
+    }
+
+    return bEnoughMemory;
+}
+
+bool URRCoreUtils::ShutDownSim(const UObject* InContextObject, uint64 InSimCompletionTimeoutInSecs)
+{
+    // END ALL SCENE INSTANCES' OPERATIONS --
+    auto* gameState = URRCoreUtils::GetGameState<ARRGameState>(InContextObject);
```
chalice-graeme: GameState objects are not persistent across level loads. If this function does not tear down the async calls before a level transition this will likely crash.

Source/RapyutaSimulationPlugins/Private/Core/RRProceduralMeshComponent.cpp
```diff
@@ -30,6 +31,16 @@ URRProceduralMeshComponent::URRProceduralMeshComponent(const FObjectInitializer&
     OnMeshCreationDone.BindUObject(Cast<ARRMeshActor>(GetOwner()), &ARRMeshActor::OnBodyComponentMeshCreationDone);
 }
 
+void URRProceduralMeshComponent::Initialize(bool bIsStaticBody, bool bInIsPhysicsEnabled)
+{
+    // CustomDepthStencilValue
+    ARRMeshActor* ownerActor = CastChecked<ARRMeshActor>(GetOwner());
+    if (ownerActor->GameMode->IsDataSynthSimType() && ownerActor->IsDataSynthEntity())
+    {
+        SetCustomDepthStencilValue(ownerActor->ActorCommon->GenerateUniqueDepthStencilValue());
```
chalice-graeme, line 41: Is there any chance ActorCommon can be null here?

Source/RapyutaSimulationPlugins/Private/Core/RRStaticMeshComponent.cpp
```diff
@@ -55,7 +57,11 @@ void URRStaticMeshComponent::Initialize(bool bInIsStationary, bool bInIsPhysicsE
     SetSimulatePhysics(bInIsPhysicsEnabled);
 
     // CustomDepthStencilValue
-    SetCustomDepthStencilValue(URRActorCommon::GenerateUniqueDepthStencilValue());
+    ARRMeshActor* ownerActor = CastChecked<ARRMeshActor>(GetOwner());
+    if (ownerActor->GameMode->IsDataSynthSimType() && ownerActor->IsDataSynthEntity())
+    {
+        SetCustomDepthStencilValue(ownerActor->ActorCommon->GenerateUniqueDepthStencilValue());
```
chalice-graeme, line 64: As above, can ActorCommon be null here?

Source/RapyutaSimulationPlugins/Private/Core/RRUObjectUtils.cpp
```diff
@@ -145,6 +149,65 @@ ARRBaseActor* URRUObjectUtils::SpawnSimActor(UWorld* InWorld,
...
+        OutCenterAndVertices.Emplace(center +
+                                     actorQuat.RotateVector(FVector((0.f == vertexNormal.X) ? -extentsLocal.X : extentsLocal.X,
+                                                                    (0.f == vertexNormal.Y) ? -extentsLocal.Y : extentsLocal.Y,
+                                                                    (0.f == vertexNormal.Z) ? -extentsLocal.Z : extentsLocal.Z)));
+    }
+#if RAPYUTA_SIM_VISUAL_DEBUG
+    if (IsInGameThread() && (nullptr == InBaseActor))
+    {
+        // Draw [InActor]'s bounding box before projecting it onto screen to get the 2D coordinates.
+        OutCenterAndVertices.RemoveAt(0);
```
chalice-graeme: Why remove this then insert at the same index rather than just replacing the value at index 0? This will do at lest two memmoves of all remaining vertices in that array.

Source/RapyutaSimulationPlugins/Public/Core/RRThreadUtils.h
```diff
@@ -82,4 +93,51 @@ class RAPYUTASIMULATIONPLUGINS_API URRThreadUtils : public UBlueprintFunctionLib
             MoveTemp(InTask),
             TUniqueFunction<void()>([InCompletionCallback = MoveTemp(InCompletionCallback)]() { InCompletionCallback(); }));
     }
+    template<typename TResult>
+    static auto AddAsyncTaskInThreadPool(FRRAsyncJob& OutAsyncJob,
```
chalice-graeme: This function returns nothing, so the return type will always default to void. I think it's better to be explicit and change it.

Source/RapyutaSimulationPlugins/Public/Core/RRThreadUtils.h
```diff
@@ -82,4 +93,51 @@ class RAPYUTASIMULATIONPLUGINS_API URRThreadUtils : public UBlueprintFunctionLib
...
+               TEXT("[%ld:%s] ASYNC JOB NUM: %d"),
+               InCurrentCaptureBatchId,
+               *OutAsyncJob.JobName,
+               OutAsyncJob.GetTasksNum());
+#endif
+        OutAsyncJob.AddAsyncTask(InCurrentCaptureBatchId, DoAsyncTaskInThread(MoveTemp(InTask), MoveTemp(InCompletionCallback)));
+    }
+
+    template<typename TResult>
+    static auto AddAsyncTaskToJob(FRRAsyncJob& OutAsyncJob,
```
chalice-graeme: Same as above.

chalice-graeme (changes requested):

Source/RapyutaSimulationPlugins/Private/Core/RRCoreUtils.cpp
Tadinu: Right, I just want to avoid a bit unwieldy static_cast<uint8> from the enum value. Given the map size as the no of file types being not so large, its overheads should be acceptable I suppose.

chalice-graeme: Two options are to make this a C-style enum, or make inline functions to do the cast if there are a lot of places this map is being accessed. A C-style array will not allocate any memory and indexing can be further inlined if the context can be statically determined. A TMap is a hashtable which is unlikely to be the same size as your constant elements and will likely change size as you add to it, meaning it adds to memory fragmentation as part of its initialization step adds all the values to it.

PR #64, opened 2022-06-15
Add skeletal turtlebot3 BP

vilkun:


Tadinu:
@vilkun It seems `Content/Robots` folder has been also updated?

yuokamoto:
Thanks you.
Could add explanation of each model directly somethwere such as readme, e.g. BallCasterSphereWheeled, StaticMeshConstrained.

and I got following error, is this ok?


https://user-images.githubusercontent.com/16463135/174468082-6b7a26cf-970d-405d-b1b5-316f2e014038.mp4

Tadinu:
@yuokamoto Fixed skmesh for all physics assets now, probably they were somehow updated due to my renaming.

yuokamoto:
Confirmed models are working. few questions

- StaticMeshConstrained and WheeeledVehicle 's BP seems has wrong scale setting. those are much larger than others
- it seems almost all of them has setting to move automatically. I guess those comes from testing purpose. We can keep those setting but maybe better to add note in documentation.

vilkun:
StaticMeshConstrained and WheeeledVehicle ', by default it's better to export STL to FBX with 0.01 scale, then import to asset and put to the scene with 0.1 scale because it's actually Rapyuta level with 0.1 scales (so less than actual world values).
Also, the robot with vehicle component with 10x scale, because STL wrongly imports to FBX with 0.001 scale (error in skinning), and we can't scale down to 0.1 this robot, because of the vehicle component (it doesn't wheel at all after changing scaling)

yuokamoto (approved):

PR #65, opened 2022-06-16
add a base class without skeletalMesh

tetorea:
Using a skeletalMesh as a root component results in problems when using collisions.
The collision needs to be set on the skeletal mesh, and thus the skeletal mesh needs to be the moving part that pick payloads. But then, the robot position appears to move in /odom

So I added the class `ARobotEmptyVehicle` which is like `ARobotVehicle` but without the skeletalMesh component. And the `ARobotVehicle` class is derived from `ARobotEmptyVehicle`, adding the skeletal mesh like before.

### DIFF
--- a/Source/RapyutaSimulationPlugins/Private/Robots/RobotVehicle.cpp
+++ b/Source/RapyutaSimulationPlugins/Private/Robots/RobotVehicle.cpp
@@ -14,15 +14,15 @@
 
 ARobotVehicle::ARobotVehicle()
 {
-    SetupDefault();
+    SetupRootSkeletal();
 }
 
 ARobotVehicle::ARobotVehicle(const FObjectInitializer& ObjectInitializer) : Super(ObjectInitializer)
 {
-    SetupDefault();
+    SetupRootSkeletal();
 }
 
-void ARobotVehicle::SetupDefault()
+void ARobotVehicle::SetupRootSkeletal()
 {
     // Generally, for sake of dynamic robot type import/creation, child components would be then created on the fly!
     // Besides, a default subobject, upon content changes, also makes the owning actor become vulnerable since one in child BP actor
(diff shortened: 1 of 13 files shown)


Source/RapyutaSimulationPlugins/Private/Robots/RobotVehicle.cpp
```diff
@@ -14,15 +14,15 @@
...
 }
 
 ARobotVehicle::ARobotVehicle(const FObjectInitializer& ObjectInitializer) : Super(ObjectInitializer)
 {
-    SetupDefault();
+    SetupRootSkeletal();
 }
 
-void ARobotVehicle::SetupDefault()
+void ARobotVehicle::SetupRootSkeletal()
```
Tadinu: I'd suggest `SetupDefault()` name be kept since it literally signify the function's meaning as to create default child objects/components that are exposed to BP users so they could config them statically.

Source/RapyutaSimulationPlugins/Public/Robots/RobotEmptyVehicle.h
```diff
@@ -0,0 +1,149 @@
...
+
+/**
+ * @brief Base RobotEmptyVehicle class. Other robot class should inherit from this class.
+ * This actor moves with #URobotVehicleMovementComponent.
+ * This actor is possessed by #ARRRobotVehicleROSController to be control from ROS2.
+ * You can find example at #ATurtlebotBurger.
+ *
+ */
+UCLASS()
+class RAPYUTASIMULATIONPLUGINS_API ARobotEmptyVehicle : public ARRBaseActor
```
Tadinu, line 36: Since this base vehicle class is not really empty as housing links/joints and a default scene component as root, I'd suggest its name be `ARRRobotBaseVehicle` to make it consistent with the naming convention as in `ARRBaseActor`.

(Regarding `ARobotVehicle, `ASimulationState`, which should be prefixed `RR` as well, however because of the complicated overheads required for child BP class changes, it is not very straightforward)

Tadinu: Since this base vehicle class is not really empty as housing links/joints and a default scene component as root, I'd suggest its name be `ARRRobotBaseVehicle` to make it consistent with the naming convention as in `ARRBaseActor`.

(Regarding `ARobotVehicle, ASimulationState`, which should be prefixed `RR` as well, however because of the complicated overheads required for child BP class changes, it is not very straightforward)

Tadinu:
Thanks for the reasonable changes, since `ARobotVehicle` is quite a sensitive class due to parenting various BP actor class, some check might be required for io_amr_UE & tb3-UE also I guess.

yuokamoto:
Cool. Please
- check io_amr_UE and turtlebot3-UE build pass
- update documentation.

Source/RapyutaSimulationPlugins/Private/Robots/RobotVehicle.cpp
tetorea: I use SetupDefault() in the parent class and it creates the default object there.
SetupRootSkeletal() only creates the skeletalMesh and sets it as the root component so I renamed it

tetorea:
ok, I'm still doing some tests but I'll test io_amr_UE and Turtlebot3 also

Source/RapyutaSimulationPlugins/Private/Robots/RobotVehicle.cpp
Tadinu: Ah right, sorry, since the function is supposed to only be called inside ctor, thus could not be made virtual, so maybe it should be `SetupDefaultRootSkeletal()` just for clarity.

tetorea:
Compilation and run tests ok for io_amr_UE and turtlebot3-UE

yuokamoto (approved):

PR #67, opened 2022-06-24
Add ARRBaseRobot + URRRobotROS2Interface

Tadinu:
* Add `ARRBaseRobot` with most common attributes: `RobotID, RobotUniqueName, RobotModelName, ROS2IntefaceClass` without `VehicleMovementComponent`, so directly inheritable by a robot arm.
* Inheritance line `ARobotVehicle -> ARobotBaseVehicle -> ARRBaseRobot`
* Prefix RR->RobotBaseVehicle
* Add `URRRobotROS2Interface`, responsible for robot's ROS2 subscription, publication
* `ARRRobotVehicleROSController`: Instantiate `URRRobotROS2Interface` on OnPossess (still wonder why we have to init robot's ROS2 communication from AIController but not right after being spawned?)
* Rem `ATurtleborROSController`, use `URRTurtlebotROS2Interface`
* Rem `BP_TurlebotROSController`, use `BP_TurtlebotROS2Interface`
* Use `UROS2Publisher::CreatePublisher()` added in https://github.com/rapyuta-robotics/rclUE/pull/53


Source/RapyutaSimulationPlugins/Private/Core/RRROS2Utils.cpp
```diff
@@ -0,0 +1,17 @@
+// Copyright 2020-2022 Rapyuta Robotics Co., Ltd.
```
yuokamoto: Do you need this class? 
isn't is simple to add constructor with same argument to ROS2 Publisher?

Source/RapyutaSimulationPlugins/Public/Core/RRROS2Utils.h
```diff
@@ -0,0 +1,48 @@
+/**
```
yuokamoto: this purely ros node, publisher, subscribe utils should be in rclUE if necessary.

yuokamoto (commented):
Please make pr as small as possible.

Tadinu:
Noted, I will try, this one should not have to include the RR prefix refactoring, which could have been done originally.

Source/RapyutaSimulationPlugins/Private/Core/RRROS2Utils.cpp
Tadinu: Ah indeed, this could be moved UROS2Publisher class

Source/RapyutaSimulationPlugins/Private/Robots/RRBaseRobot.cpp
```diff
@@ -0,0 +1,87 @@
...
+    SetupDefault();
+}
+
+void ARRBaseRobot::SetupDefault()
+{
+    // Generally, for sake of dynamic robot type import/creation, child components would be then created on the fly!
+    // Besides, a default subobject, upon content changes, also makes the owning actor become vulnerable since one in child BP actor
+    // classes will automatically get invalidated.
+
+    RootComponent = CreateDefaultSubobject<USceneComponent>(TEXT("SceneComp"));
```
yuokamoto: need to be sync with torea's pr as well

Source/RapyutaSimulationPlugins/Public/Robots/RRRobotROS2Interface.h
```diff
@@ -0,0 +1,123 @@
...
+    UPROPERTY(BlueprintReadWrite)
+    FString CmdVelTopicName = TEXT("cmd_vel");
+
+    //! Joint control command topic. If empty is given, subscriber will not be initiated.
+    UPROPERTY(BlueprintReadWrite)
+    FString JointsCmdTopicName = TEXT("joint_states");
+
+protected:
+    /**
+     * @brief Create a ROS2 publisher
```
yuokamoto, line 113: argument order is better to be Out*  arg is after all In* args

Source/RapyutaSimulationPlugins/Public/Robots/RRRobotROS2Interface.h
```diff
@@ -0,0 +1,123 @@
...
+     *
+     * @param InPawn
+     */
+    void InitRobotROS2Node(ARobotVehicle* InRobot);
+
+    UPROPERTY(Transient, BlueprintReadWrite)
+    URRROS2OdomPublisher* OdomPublisher = nullptr;
+
+    /**
+     * @brief Initialize non sensor publishes such as odometry.
```
yuokamoto: better to have more explanation for initpublishers, e.g. 
- this method is expected to be override for each robot
- base class has odom publisher initialization as a example for mobile robot
-

Source/RapyutaSimulationPlugins/Public/Robots/RRRobotROS2Interface.h
```diff
@@ -0,0 +1,123 @@
...
+     *
+     */
+    UFUNCTION()
+    virtual void StopPublishers();
+
+    /**
+     * @brief Initialize subscriptions
+     *
+     */
+    UFUNCTION()
```
yuokamoto, line 75: better to have more explanation for init subscripsions, e.g. 
- this method is expected to be override for each robot
- base class has cmd_vel and joint state subscription

yuokamoto (commented):
maybe better to have 2 seprate ros interface for base robot(arm) and mobilerobot, e.g.
- BaseRoboTRosInterface which has joint cmd subscriber and joint state publisher.
- MobileRobotROSInterface which has odom publisher and cmd_vel subscriber.child class of baserobotrosinterfaces.

but can be don in separate following pr.

Tadinu:
Yeah, in fact mobility & articulation are two orthogonal features of a generic robot, meaning either of those could be the base one. Like for an arm-attached agv, when it stays in one place, it is a manipulator & when it moves it is a vehicle.
Thus, a barebone RobotROS2Interface should actually be composed of just basic publishing/subscribing setup, and leave specific data (like odom or joint_state) particulars to child classes (like MobileROS2Interface + ArmROS2Inteface), which could be done upon legit needs.

yuokamoto (commented):
Please solve conflict and test build with jenkins.
others looks good to me.

tetorea (approved):
works with my project

Tadinu:
[io_amr_UE's PR](https://github.com/rapyuta-robotics/io_amr_UE/pull/88) build [completed](https://rr-ss-ci.japaneast.cloudapp.azure.com/job/io_amr_UE%20github%20builder/job/PR-88/5/consoleFull#:~:text=BUILD%20COMMAND%20COMPLETED)

PR #69, opened 2022-06-27
Collision fix2

tetorea:
1) Only the root component of the robot BP is used to handle collisions.
When a robot uses a moving part that can collide with environment objects, we may need to keep the robot reference pose immobile while the skeletal mesh moves.
For such cases, I added 2 RootOffset variables, one for /odom and one for /ue_ros/model_state, that can be modified to correct the reference pose of the robot to compensate for the skeletal mesh local motion.

2) adding a variable to ignore warnings when some links are not available

### DIFF
--- a/Source/RapyutaSimulationPlugins/Private/Drives/RobotVehicleMovementComponent.cpp
+++ b/Source/RapyutaSimulationPlugins/Private/Drives/RobotVehicleMovementComponent.cpp
@@ -93,8 +93,7 @@ void URobotVehicleMovementComponent::UpdateMovement(float InDeltaTime)
             else
             {
                 // very basic robot falling
-                PawnOwner->AddActorWorldOffset(
-                    FVector(0., 0., -FallingSpeed * InDeltaTime), true, &hit, ETeleportType::None);
+                PawnOwner->AddActorWorldOffset(FVector(0., 0., -FallingSpeed * InDeltaTime), true, &hit, ETeleportType::None);
             }
         }
         else
@@ -159,11 +158,10 @@ void URobotVehicleMovementComponent::UpdateMovement(float InDeltaTime)
                 // Moves the robot up or down, depending on impact position
                 float minDistance = *Algo::MinElement(contactsDistance);
                 minDistance -= RayOffsetUp;
-
                 minDistance = FMath::Min(minDistance, FallingSpeed * InDeltaTime);
 
                 FVector heightVariation = {0., 0., -minDistance};
-                PawnOwner->AddActorWorldOffset(heightVariation, true, &hit, ETeleportType::None);
+                PawnOwner->AddActorWorldOffset(heightVariation, false, &hit, ETeleportType::None);
             }
         }
     }
@@ -199,6 +197,7 @@ void URobotVehicleMovementComponent::InitOdom()
     OdomData.pose_pose_orientation = InitialTransform.GetRotation();
 
     PreviousTransform = InitialTransform;
+    PreviousNoisyTransform = InitialTransform;
 
     // todo temporary hardcoded
     OdomData.pose_covariance.Init(0, 36);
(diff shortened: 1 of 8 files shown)


Tadinu:
There seem to be files from `Contents`, which have been added to lfs [here](https://github.com/rapyuta-robotics/RapyutaSimulationPlugins/blob/devel/.gitattributes).
Could you try reverting the commit & pulling from lfs then confirm by `git lfs checkout`, which should show no updated asset files.

Source/RapyutaSimulationPlugins/Private/Drives/RobotVehicleMovementComponent.cpp
```diff
@@ -159,11 +159,10 @@ void URobotVehicleMovementComponent::UpdateMovement(float InDeltaTime)
                 // Moves the robot up or down, depending on impact position
                 float minDistance = *Algo::MinElement(contactsDistance);
                 minDistance -= RayOffsetUp;
-
                 minDistance = FMath::Min(minDistance, FallingSpeed * InDeltaTime);
 
                 FVector heightVariation = {0., 0., -minDistance};
-                PawnOwner->AddActorWorldOffset(heightVariation, true, &hit, ETeleportType::None);
+                PawnOwner->AddActorWorldOffset(heightVariation, false, &hit, ETeleportType::None);
```
yuokamoto, line 164: any reason to make it false?

Source/RapyutaSimulationPlugins/Private/Drives/RobotVehicleMovementComponent.cpp
```diff
@@ -324,16 +328,15 @@ void URobotVehicleMovementComponent::InitMovementComponent()
     if (bIsFloorHit)
     {
         MinDistanceToFloor = hitResult.Distance;
-        UE_LOG(LogTemp,
-               Warning,
-               TEXT("URobotVehicleMovementComponent::InitMovementComponent - Min Distance To Floor = %f"),
-               MinDistanceToFloor);
     }
+    UE_LOG(LogTemp,
```
yuokamoto: Does this woring output every tick?

Source/RapyutaSimulationPlugins/Private/Drives/RobotVehicleMovementComponent.cpp
tetorea: Here I sometimes got collisions between floor and robot where robot was visibly few centimeters above the floor, although `show collision` terminal command did not show any collisions.
At this point, collision checks between contact points and floor has already been computed so not sweeping is possible. But of course, there are possibilities that floor concavities collide with the robot body parts far from contact points then. In current project, that does not happen in normal cases.

Source/RapyutaSimulationPlugins/Private/Drives/RobotVehicleMovementComponent.cpp
tetorea: Only once at the beginning. This function is called in the constructor.

tetorea:
> There seem to be files from `Contents`, which have been added to lfs [here](https://github.com/rapyuta-robotics/RapyutaSimulationPlugins/blob/devel/.gitattributes). Could you try reverting the commit & pulling from lfs then confirm by `git lfs checkout`, which should show no updated asset files.

Done, but is there a way to check I've done it correctly?

Tadinu:
Yeah, after pulling from lfs, those real assets will be downloaded from lfs server to your PC, overriding pointers also, then by `git lfs checkout`, it should show no updates. Anw, they have been excluded from changed files now👍🏼

Tadinu (approved):
Yeah, after pulling from lfs, those real assets will be downloaded from lfs server to your PC, overriding pointers also, then by `git lfs checkout`, it should show no updates. Anw, they have been excluded from changed files now👍🏼
Thanks, could you update `LogTemp` -> `LogRapyutaCore` for consistency?

Tadinu:
Yeah, after pulling from lfs, those real assets will be downloaded from lfs server to your PC, overriding pointers also, then by `git lfs checkout`, it should show no updates. Anw, they have been excluded from changed files now👍🏼
Thanks, could you update `LogTemp` -> `LogRapyutaCore` for consistency?

yuokamoto (approved):

PR #73, opened 2022-07-12
SimulationState Add SpawnEntity() used for SpawnEntitySrv + SpawnEntityListSrv

Tadinu:
Group duplicated src of single entity spawning into 
```
AActor* SpawnEntity(const FROSSpawnEntityRequest& InROSSpawnRequest, const TSubclassOf<AActor>& InEntityClass, const FTransform& InEntityTransform);
```

### DIFF
--- a/Source/RapyutaSimulationPlugins/Private/Tools/SimulationState.cpp
+++ b/Source/RapyutaSimulationPlugins/Private/Tools/SimulationState.cpp
@@ -18,6 +18,7 @@
 // RapyutaSimulationPlugins
 #include "Core/RRActorCommon.h"
 #include "Core/RRConversionUtils.h"
+#include "Core/RRUObjectUtils.h"
 #include "Tools/ROS2Spawnable.h"
 
 // Sets default values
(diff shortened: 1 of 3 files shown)


Source/RapyutaSimulationPlugins/Public/Tools/SimulationState.h
```diff
@@ -56,32 +58,33 @@ class RAPYUTASIMULATIONPLUGINS_API ASimulationState : public AActor
...
      */
     ASimulationState();
 
 public:
     /**
      * @brief Start ROSservices.
-     * 
-     * @param InROS2Node @sa [ROS2Node](https://rclue.readthedocs.io/en/devel/doxygen_generated/html/d6/dcb/class_a_r_o_s2_node.html)
+     *
+     * @param InROS2Node @sa
```
yuokamoto: can you  avoid change line after @sa or change line before @sa

Tadinu: Yeah, unclear why somehow clang-format does it.

Source/RapyutaSimulationPlugins/Private/Tools/SimulationState.cpp
```diff
@@ -240,32 +242,70 @@ void ASimulationState::AttachSrv(UROS2GenericSrv* Service)
...
+            FTransform relativeTransf =
+                URRConversionUtils::TransformROSToUE(FTransform(spawnEntityRequest.StatePoseOrientation, relLocation));
             FTransform worldTransf;
-            URRGeneralUtils::GetWorldTransform(
-                Request.StateReferenceFrame,
-                Entities.Contains(Request.StateReferenceFrame) ? Entities[Request.StateReferenceFrame] : nullptr,
-                relativeTransf,
-                worldTransf);
+            URRGeneralUtils::GetWorldTransform(spawnEntityRequest.StateReferenceFrame,
+                                               Entities.FindRef(spawnEntityRequest.StateReferenceFrame),
```
yuokamoto: Does FindRef return null if entities do not have item?

Tadinu: The name could be abit mis leading, since it actually returns a value, not a reference.
As defined [here](https://github.com/EpicGames/UnrealEngine/blob/release/Engine/Source/Runtime/Core/Public/Containers/Map.h#L673), it returns default value of `AActor*`, which should be [zero](https://godbolt.org/z/sn8TqGeah).
A quick check could help verify it:
```
    TMap<const TCHAR*, AActor*> testMap = {{TEXT("a"), this}};
    verify(nullptr == testMap.FindRef(TEXT("b")));
```

Source/RapyutaSimulationPlugins/Private/Tools/SimulationState.cpp
```diff
@@ -41,7 +43,7 @@ void ASimulationState::Init(AROS2Node* InROS2Node)
     SetEntityStateSrvCallback.BindDynamic(this, &ASimulationState::SetEntityStateSrv);
     AttachSrvCallback.BindDynamic(this, &ASimulationState::AttachSrv);
     SpawnEntitySrvCallback.BindDynamic(this, &ASimulationState::SpawnEntitySrv);
-    SpawnEntitiesSrvCallback.BindDynamic(this, &ASimulationState::SpawnEntitiesSrv);
+    SpawnEntitiesSrvCallback.BindDynamic(this, &ASimulationState::SpawnEntityListSrv);
```
tetorea: I would prefer to keep similar name everywhere, either Entities or EntityList.

yuokamoto (approved):
lgtm

Source/RapyutaSimulationPlugins/Private/Tools/SimulationState.cpp
```diff
@@ -240,32 +242,70 @@ void ASimulationState::AttachSrv(UROS2GenericSrv* Service)
     AttachService->SetResponse(Response);
 }
 
+AActor* ASimulationState::SpawnEntity(const FROSSpawnEntityRequest& InROSSpawnRequest,
+                                      const TSubclassOf<AActor>& InEntityClass,
+                                      const FTransform& InEntityTransform)
+{
+    // SpawnActorDeferred to set parameters beforehand
+    AActor* newEntity = GetWorld()->SpawnActorDeferred<AActor>(InEntityClass, InEntityTransform);
+    if (nullptr == newEntity)
```
tetorea: Just a detail but I find it more readable to put the variable tested before the value tested `newEntity == nullptr`.
I think this was an optimization trick before but recent compilers generate the same assembly code now

Source/RapyutaSimulationPlugins/Private/Tools/SimulationState.cpp
```diff
@@ -240,32 +242,70 @@ void ASimulationState::AttachSrv(UROS2GenericSrv* Service)
...
+    AddEntity(newEntity);
+    return newEntity;
+}
+
 void ASimulationState::SpawnEntitySrv(UROS2GenericSrv* Service)
 {
     UROS2SpawnEntitySrv* SpawnEntityService = Cast<UROS2SpawnEntitySrv>(Service);
 
-    FROSSpawnEntityRequest Request;
-    SpawnEntityService->GetRequest(Request);
```
tetorea, line 248: I would keep the old names :
- Request and Response are used in all other functions
- from the context, we know this is the SpawnEntityRequest
- it becomes longer to read and write. less verbosity (and without ambiguity) => faster to read and process the function

Source/RapyutaSimulationPlugins/Private/Tools/SimulationState.cpp
```diff
@@ -240,32 +242,70 @@ void ASimulationState::AttachSrv(UROS2GenericSrv* Service)
...
+    FROSSpawnEntityRequest spawnEntityRequest;
+    SpawnEntityService->GetRequest(spawnEntityRequest);
 
-    FROSSpawnEntityResponse Response;
-    Response.bSuccess = CheckSpawnableEntity(Request.Xml, false) && CheckEntity(Request.StateReferenceFrame, true);
+    FROSSpawnEntityResponse spawnEntityResponse;
```
tetorea: same as above

Source/RapyutaSimulationPlugins/Private/Tools/SimulationState.cpp
```diff
@@ -274,119 +314,94 @@ void ASimulationState::SpawnEntitySrv(UROS2GenericSrv* Service)
...
-    FROSSpawnEntitiesRequest request;
-    spawnEntitiesService->GetRequest(request);
+    FROSSpawnEntitiesRequest entityListRequest;
+    spawnEntitiesService->GetRequest(entityListRequest);
 
-    int numEntitySpawned = 0;
-    FString statusMessage = "";
-    for (uint i = 0; i < request.NameList.Num(); ++i)
+    int32 spawnedEntitiesNum = 0;
+    FString finalStatusMessage;
```
tetorea: I would also have kept the old variable name : it's shorter and there is no other status message used in the function anyway

Source/RapyutaSimulationPlugins/Private/Tools/SimulationState.cpp
Tadinu: Yeah, it is mostly a preference, https://en.wikipedia.org/wiki/Yoda_conditions
To me either way would be similarly readable.

Source/RapyutaSimulationPlugins/Private/Tools/SimulationState.cpp
Tadinu: I typically avoid ambiguous var name regardless of the context or its class type (like if you have more than 2 ones, the sources would be come a bit cryptic). But in this case, since it's not a long function so you are right👍🏼

Source/RapyutaSimulationPlugins/Private/Tools/SimulationState.cpp
Tadinu: Yeah, it is mostly a personal preference, https://en.wikipedia.org/wiki/Yoda_conditions
To me either way would be similarly readable.

tetorea (approved):

PR #77, opened 2022-08-03
ARRNetworkPlayerController::FindPawnToPossess() Use APawn::GetPlayerState()

yuokamoto:
- add prefix to log to differentiate client and server to test in editor
- introduce isPossed member to base robot since APawn::GetController do not work in multiplaeyer


Source/RapyutaSimulationPlugins/Private/Core/RRNetworkPlayerController.cpp
```diff
@@ -88,9 +88,10 @@ void ARRNetworkPlayerController::InitSimStateClientROS2()
     // Init SimStateClient's [ROS2Node] & [ClockPublisher]
     UWorld* currentWorld = GetWorld();
     SimStateClientROS2Node = currentWorld->SpawnActor<AROS2Node>();
+    SimStateClientROS2Node->AttachToActor(this, FAttachmentTransformRules::KeepRelativeTransform);
```
Tadinu, line 91: We may need to switch `AROS2Node` to an Actor component or just UObject then since it seems to have no Actor/Pawn functionalitites.

Tadinu, line 91: We may need to switch `AROS2Node` to an Actor component or just UObject then since it seems to have no Actor/Pawn functionalities.

Source/RapyutaSimulationPlugins/Private/Core/RRNetworkPlayerController.cpp
```diff
@@ -183,10 +188,15 @@ APawn* ARRNetworkPlayerController::FindPawnToPossess()
...
-        if (IsValid(matchingPawn) && (nullptr == matchingPawn->GetController()) && (GetPawn() != matchingPawn))
+        ARRBaseRobot* robot = Cast<ARRBaseRobot>(entity);
+        if (IsValid(robot) && !robot->isPossessed && (GetPawn() != robot))
         {
             break;
         }
+        else
+        {
+            matchingPawn = nullptr;
+        }
```
Tadinu: `matchingPawn = Cast<APawn>(entity)` could be moved to before `break`, so we would not need to reassign it to `nullptr`?

Source/RapyutaSimulationPlugins/Private/Tools/RRROS2SimulationStateClient.cpp
```diff
@@ -186,8 +191,10 @@ void URRROS2SimulationStateClient::AttachSrv(UROS2GenericSrv* InService)
     {
         UE_LOG(LogRapyutaCore,
                Warning,
-               TEXT("Entity %s and/or %s not exit or not under SimulationState Actor control. Please call AddEntity to make Actors "
+               TEXT("[%s] Entity %s and/or %s not exit or not under SimulationState Actor control. Please call AddEntity to make "
+                    "Actors "
```
Tadinu:
```suggestion
               TEXT("[%s] Entity %s and/or %s not exit or not under SimulationState Actor control. Please call ServerAddEntity to make "
                    "Actors "
```

Source/RapyutaSimulationPlugins/Private/Tools/SimulationState.cpp
```diff
@@ -377,6 +377,13 @@ AActor* ASimulationState::ServerSpawnEntity(const FROSSpawnEntityRequest& InRequ
 
             // Spawn entity
             newEntity = ServerSpawnEntity(InRequest, SpawnableEntityTypes[entityModelName], worldTransf);
+            if (nullptr == newEntity)
+            {
+                // response.bSuccess = false;
+                // response.StatusMessage =
+                //     FString::Printf(TEXT("[%s] Failed to spawn entity named %s, probably out collision!"), *GetName(),
+                //     *entityName);
```
Tadinu, line 386:
```suggestion
                UE_LOG(LogRapyutaCore, Error, TEXT("[ASimulationState] Failed to spawn entity named %s, probably out collision!"), *entityName);
```

Source/RapyutaSimulationPlugins/Public/Robots/RRBaseRobot.h
```diff
@@ -57,6 +57,22 @@ class RAPYUTASIMULATIONPLUGINS_API ARRBaseRobot : public ARRBaseActor
     //! Robot's ROS2 Interface
     UPROPERTY(Replicated)
     URRRobotROS2Interface* ROS2Interface = nullptr;
+
+    //! is this robot is possed or not. Defaults method's do not works with multiplayer.
+    UPROPERTY(Replicated)
+    bool isPossessed = false;
```
Tadinu: `b` prefix for boolean
```suggestion
    //! Whether this robot is possessed or not. `GetController()` do not work with multiplayer.
    UPROPERTY(Replicated)
    bool bIsPossessed = false;
```
Just to confirm, [here](https://github.com/EpicGames/UnrealEngine/blob/release/Engine/Source/Runtime/Engine/Private/Pawn.cpp#L559) it seems `APawn::Controller` does get replicated?
Also `APawn::GetLifetimeReplicatedProps()` has it fetched. @larryng-rapyuta

Source/RapyutaSimulationPlugins/Private/Core/RRNetworkPlayerController.cpp
yuokamoto: https://www.wrike.com/workspace.htm?acc=1794223#folder/600436429/list?filters=status%3Dactive&sidePanelItemId=942107253&sortOrder=5&spaceId=364052400&viewId=12663565

created task

Source/RapyutaSimulationPlugins/Public/Robots/RRBaseRobot.h
yuokamoto: I did experienced  like this just inside FindPawnToPossess, and result is 0 1 1 just after posses by one of networkplayer
```
            UE_LOG(LogRapyutaCore,
                   Error,
                   TEXT("Player [%s] %d %d %d"),
                   *PlayerName,
                   !robot->IsPossessed(),
                   nullptr == robot->GetController(),
                   nullptr == robot->Controller);

File: Source/RapyutaSimulationPlugins/Private/Core/RRNetworkPlayerController.cpp (hunk 1 above)
Comment (larryng-rapyuta) [reply Source/RapyutaSimulationPlugins/Private/Core/RRNetworkPlayerController.cpp:91]: That might be a good idea to make it a component since we always seem to have an actor that defines all the services/msgs and then its initialized in the ROS2Node, could be easier if the ROS2node was a component and just picked it up from the actor/object it was attached it

Comment (larryng-rapyuta) [inline Source/RapyutaSimulationPlugins/Private/Core/RRNetworkPlayerController.cpp:91]: If we need to attach the SimStateClient Ros2node to something, shouldn't we be attaching it to the SimulationStateClient instead of the Player controller?

**Review by larryng-rapyuta — APPROVED** [review]

File: Source/RapyutaSimulationPlugins/Public/Robots/RRBaseRobot.h (hunk 5 above)
Comment (Tadinu) [reply Source/RapyutaSimulationPlugins/Public/Robots/RRBaseRobot.h]: @larryng-rapyuta?

File: Source/RapyutaSimulationPlugins/Public/Robots/RRBaseRobot.h [hunk 6]
```diff
@@ -182,9 +167,31 @@ class RAPYUTASIMULATIONPLUGINS_API ARRBaseRobot : public ARRBaseActor
...
+        return bIsPossessed;
+    }
+
 protected:
     /**
      * @brief Instantiate default child components
      */
     virtual void PostInitializeComponents() override;
+
+    //! is this robot is possed or not. Defaults method's do not works with multiplayer.
```
Tadinu: `Default methods` seems to be a bit vague, should it be `GetController()`?
So if `APawn::Controller` does get replicated, why didn't it show up in other clients tho? 🤔

yuokamoto: I'm not sure but I tried methods in the comment but do not work.
Controller replication seems not simple?
https://github.com/EpicGames/UnrealEngine/blob/d9d435c9c280b99a6c679b517adedd3f4b02cfd7/Engine/Source/Runtime/Engine/Private/Pawn.cpp#L559

Source/RapyutaSimulationPlugins/Public/Robots/RRBaseRobot.h
Tadinu: Yeah, it looks like controller is not really replicated. This may give some insight
https://forums.unrealengine.com/t/in-multiplayer-mode-cant-get-other-characters-player-id-of-other-clients-in-current-client/578498/7

They recommend `PlayerState` instead, could you try using it?
https://github.com/EpicGames/UnrealEngine/blob/release/Engine/Source/Runtime/Engine/Classes/GameFramework/Pawn.h#L153

yuokamoto: Thanks, it works

Tadinu (approved):

PR #78, opened 2022-08-05
Fix ros2 interface's publisher init

Tadinu:
* URRRobotROS2Interface::CreatePublisher() Add QoS param to init publisher frequent callback
* Add ARRCrowdROSController, inheriting ARRCrowdAIController, to init possessed robot's ROS2Interface, used [here](https://github.com/rapyuta-robotics/RapyutaSimulationInternal/blob/SOOTBALL_SIM/Source/RapyutaSimInternal/Private/Sootball/RRSootball.cpp#L47)
* ARRRobotROS2Interface add virtual SetupROSParams(), overridable in child classes (eg with bPublishOdom off if not used)


Source/RapyutaSimulationPlugins/Public/Core/RRCrowdAIController.h
```diff
@@ -10,6 +10,13 @@
 
 #include "RRCrowdAIController.generated.h"
 
+/**
+ * @brief Base Crowd AI Controller for Character-based classes, utilizing UCrowdFollowingComponent
```
yuokamoto: pawn-based

Source/RapyutaSimulationPlugins/Public/Core/RRCrowdROSController.h
```diff
@@ -0,0 +1,37 @@
...
+#include "Core/RRCrowdAIController.h"
+
+#include "RRCrowdROSController.generated.h"
+
+class URRRobotROS2Interface;
+
+/**
+ * @brief Base Crowd ROS Controller for ARRBaseRobot descedent classes, utilizing UCrowdFollowingComponent
+ */
+UCLASS()
```
yuokamoto, line 18: Do we need RRCrowdAIController which has only different constructor from AIController?
Can't we merge RRCrowdAIController and RRCrowdROSController

Tadinu: Yeah I thought about it also, but as suggested by the name, CrowAIController should be for common pawns or robots not using ROS only.

Tadinu: Yeah I thought about it also, but as suggested by the name, CrowAIController should be only for common pawns or robots not using ROS.

yuokamoto (approved):

tetorea (approved):

PR #80, opened 2022-08-12
Add RRRobotVehiclePoseSensorManager

Tadinu:


Source/RapyutaSimulationPlugins/Private/Robots/RRRobotPoseSensorManager.cpp
```diff
@@ -0,0 +1,124 @@
+// Copyright 2020-2022 Rapyuta Robotics Co., Ltd.
+
+#include "Robots/RRRobotPoseSensorManager.h"
```
yuokamoto: should be in the sensors directorly

yuokamoto (approved):

PR #81, opened 2022-08-15
add random spawn test script

yuokamoto:
add test client script and docker 
- randomly spawn turtlebot
- random cmd_vel is sent to turtlebot. cmd_vel is updated for every 10s

maybe better to move rr_sim_tests to UE_tools or other repo.

### DIFF
--- a/Source/RapyutaSimulationPlugins/Private/Core/RRNetworkPlayerController.cpp
+++ b/Source/RapyutaSimulationPlugins/Private/Core/RRNetworkPlayerController.cpp
@@ -247,11 +247,6 @@ void ARRNetworkPlayerController::ClientInitPawn_Implementation(AActor* InActor)
         robot->CreateROS2Interface();
         robot->ROS2Interface->Initialize(robot);
     }
-
-    if (ARRRobotBaseVehicle* robotVehicle = Cast<ARRRobotBaseVehicle>(InActor))
-    {
-        robotVehicle->InitMoveComponent();
-    }
 }
 
 void ARRNetworkPlayerController::BeginPlay()
--- a/.gitignore
+++ b/.gitignore
@@ -47,7 +47,7 @@ Intermediate/*
 *_BuiltData.uasset
 
 # External test intermediate data
-ExternalTest/RapyutaSimulationPlugins_TESTS
+ExternalTest/RRSIM_TESTS
 ExternalTest/rr_sim_tests/rr_sim_tests/__pycache__
 ExternalTest/rr_sim_tests/rr_sim_tests/utils/__pycache__
 
--- a/ExternalTest/rr_sim_tests/docker/Dockerfile
+++ b/ExternalTest/rr_sim_tests/docker/Dockerfile
@@ -1,0 +1,37 @@
+FROM ros:foxy as builder
+
+# Other settings
+SHELL ["/bin/bash", "-c"]
+RUN apt update 
+
+#ROS2 
+## setup
+ENV ROS2_WS=/home/ros2_ws
+RUN mkdir -p ${ROS2_WS}/src/
+
+## clone
+WORKDIR ${ROS2_WS}/src
+RUN git clone https://github.com/rapyuta-robotics/UE_msgs.git "ue_msgs"
+COPY . rr_sim_tests/
+
+## build
+WORKDIR ${ROS2_WS}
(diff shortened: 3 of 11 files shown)


Source/RapyutaSimulationPlugins/Private/Core/RRNetworkPlayerController.cpp
```diff
@@ -247,11 +247,6 @@ void ARRNetworkPlayerController::ClientInitPawn_Implementation(AActor* InActor)
         robot->CreateROS2Interface();
         robot->ROS2Interface->Initialize(robot);
     }
-
```
yuokamoto, line 250: I needed to disable this to avoid initialize twice.
@Tadinu @larryng-rapyuta are those necessary?

Tadinu: It looks like `URobotVehicleMovementComponent` itself is not replicated (not clear whether it needs), thus comes its instantiation here

Tadinu (approved):

Source/RapyutaSimulationPlugins/Private/Core/RRNetworkPlayerController.cpp
yuokamoto: initmovecomponent is called at postinitialization method
https://github.com/rapyuta-robotics/RapyutaSimulationPlugins/blob/fe538bfcf873424f2e361954309aacfae4d7cd86/Source/RapyutaSimulationPlugins/Private/Robots/RRRobotBaseVehicle.cpp#L43

if you initialize here, it will initialize twice I think.

Tadinu: Afaik, a robot is spawned at Server side, being initialized & get replicated down to all clients with _only its ActorComponents & members that are setup for replication_.
If MovementComponent is not setup for replicated, the robot at client side would not have it I suppose.

yuokamoto: it is set as replicated here
https://github.com/rapyuta-robotics/RapyutaSimulationPlugins/blob/fe538bfcf873424f2e361954309aacfae4d7cd86/Source/RapyutaSimulationPlugins/Public/Robots/RRRobotBaseVehicle.h#L64

Tadinu: Ah that's only to replicate the pointer to its instance, which seems to appear as null on client side in my earlier try. Basically, to get an Actor component instance itself replicated, we need to invoke `SetIsReplicated(true)` like for [ROS2SimStateClient](https://github.com/rapyuta-robotics/RapyutaSimulationPlugins/blob/devel/Source/RapyutaSimulationPlugins/Private/Tools/RRROS2SimulationStateClient.cpp#L32).
But if you see it's already valid on client side, so probably meaning it's been invoked somewhere in parent classes.

Tadinu: Ah that's only to replicate the pointer to its instance, which seems to appear as null on client side in my earlier try with ROS2SimStateClient. Basically, to get an Actor component instance itself replicated, we need to invoke `SetIsReplicated(true)` like for [ROS2SimStateClient](https://github.com/rapyuta-robotics/RapyutaSimulationPlugins/blob/devel/Source/RapyutaSimulationPlugins/Private/Tools/RRROS2SimulationStateClient.cpp#L32).
But if you see it's already valid on client side, so probably meaning it's been invoked somewhere in parent classes.

yuokamoto: got it any way, need more update later
- spawn movecomponent only on client side since currently client has authority on position
- add option to spawn movecomponent on both side to support server authority as well.

Tadinu: Yeah, so I think for now it should be added to `ARRRobotBaseVehicle::PostInitializeComponents()`
```
if (IsNetMode(NM_Standalone))
{
   InitMoveComponent();
}
```
similar to ARRBaseRobot:: with ROS2Interface, which should be created by default for each robot in standalone server only.

PR #83, opened 2022-08-21
multiple robot in one client

yuokamoto:
Support multiple robot in one client.
- Use PlayerId to know which client send  the spawn request.
- ROS2Node is created only in the client which receive the spawn request  
- Since non-player pawn can't call rpc, call rpc via networkplayer controller. RRRobotBaseVehicle::ServerSet*Vel -> RRNetworkPlayerController::ServerSet*Vel.
- Since RRBaseRobot is non-player pawn, robot can't us RPC. Use replication to trigger some logic
- Add namespace to simstate ROS2Node in client to test with editor.


yuokamoto:
todo:
- test in cloud
- test with other robot

ExternalTest/rr_sim_tests/rr_sim_tests/test_random_spawn.py
```diff
@@ -43,14 +45,15 @@ def main(args=None):
...
     # sanity check
     if robot_name is None:
         print('You need to provide robot name with `--ros-args -p robot_name:=<name>`')
         return
 
-    is_robot_spawned, _ = wait_for_spawned_entity(robot_name, 10.0)
-    if is_robot_spawned
+    is_robot_spawned, _ = wait_for_spawned_entity(robot_name, service_namespace, 10.0)
+    if is_robot_spawned:
         print(robot_name + 'is already exists.')
```
Tadinu:
```suggestion
        print(robot_name + ' already exists.')
```

ExternalTest/rr_sim_tests/rr_sim_tests/test_random_spawn.py
```diff
@@ -70,12 +73,12 @@ def main(args=None):
...
-        is_srv_available = spawn_robot(robot_model, robot_name, robot_namespace, reference_frame, robot_pose)
+        is_srv_available = spawn_robot(robot_model, robot_name, robot_namespace, reference_frame, robot_pose, service_namespace)
         if is_srv_available:
             count = 0
             while count < 20:
                 time.sleep(1)
-                is_robot_spawned, _ = wait_for_spawned_entity(robot_name, 10.0)
+                is_robot_spawned, _ = wait_for_spawned_entity(robot_name, service_namespace, 10.0)
                 if is_robot_spawned:
                     res = False
```
Tadinu: `res` as `False` for a successful spawn may not be intuitive & it seems meant to only break out of while loop, so could it be `while True` & `break`?

ExternalTest/rr_sim_tests/rr_sim_tests/utils/utils.py
```diff
@@ -23,11 +24,11 @@
 
 SERVICE_NAME_SPAWN_ENTITY = 'SpawnEntity'
 
-def spawn_robot(in_robot_model, in_robot_name, in_robot_namespace, in_robot_ref_frame, in_robot_pose, in_timeout=5.0):
+def spawn_robot(in_robot_model, in_robot_name, in_robot_namespace, in_robot_ref_frame, in_robot_pose, service_namespace='', in_timeout=5.0):
     assert(len(in_robot_model) > 0)
     assert(len(in_robot_name) > 0)
     node = rclpy.create_node(f'spawn_{in_robot_name}')
-    cli = wait_for_service(node, SpawnEntity, SERVICE_NAME_SPAWN_ENTITY)
+    cli = wait_for_service(node, SpawnEntity, os.path.join(service_namespace, SERVICE_NAME_SPAWN_ENTITY))
```
Tadinu: `os.path.join` in Windows would result in `\`, which is accepted in a ros namespace?

ExternalTest/rr_sim_tests/rr_sim_tests/utils/wait_for_spawned_entity.py
```diff
@@ -9,10 +10,10 @@
 
 SERVICE_NAME_GET_ENTITY_STATE = 'GetEntityState'
 
-def wait_for_spawned_entity(in_entity_name, in_timeout=10.0):
+def wait_for_spawned_entity(in_entity_name, service_namespace='', in_timeout=10.0):
     print(f'Waiting for entity [{in_entity_name}]...')
     node = rclpy.create_node(f'wait_for_{in_entity_name}')
-    cli = wait_for_service(node, GetEntityState, SERVICE_NAME_GET_ENTITY_STATE, in_timeout)
+    cli = wait_for_service(node,GetEntityState,  os.path.join(service_namespace, SERVICE_NAME_GET_ENTITY_STATE), in_timeout)
```
Tadinu: Same as above

Source/RapyutaSimulationPlugins/Public/Robots/RRBaseRobot.h
```diff
@@ -55,14 +71,59 @@ class RAPYUTASIMULATIONPLUGINS_API ARRBaseRobot : public ARRBaseActor
...
     URRRobotROS2Interface* ROS2Interface = nullptr;
+    
+    UFUNCTION(BlueprintCallable)
+    virtual void OnRep_ROS2Interface();
+
+    /**
+     * @brief Flag to start/stop ROS2Interfaces. Since RPC can't be used, use replication to trigger initialization.
+     */   
+    UPROPERTY(VisibleAnywhere, Replicated, ReplicatedUsing = OnRep_StartStopROS2Interface)
+    bool StartStopROS2Interface = false;
```
Tadinu:
```suggestion
    bool bStartStopROS2Interface = false;
```
Guess we had this noted various times before for `b` prefix with boolean as to be easily recognized in a check versus a pointer

Source/RapyutaSimulationPlugins/Private/Core/RRNetworkPlayerController.cpp
```diff
@@ -34,28 +33,21 @@ ARRNetworkPlayerController::ARRNetworkPlayerController()
...
     UpdateLocalClock(DeltaSeconds);
 }
 
 void ARRNetworkPlayerController::GetLifetimeReplicatedProps(TArray<FLifetimeProperty>& OutLifetimeProps) const
 {
     Super::GetLifetimeReplicatedProps(OutLifetimeProps);
-    DOREPLIFETIME(ARRNetworkPlayerController, SimStateClientROS2Node);
-    DOREPLIFETIME(ARRNetworkPlayerController, SimStateClientClockPublisher);
+    // DOREPLIFETIME(ARRNetworkPlayerController, SimStateClientROS2Node);
+    // DOREPLIFETIME(ARRNetworkPlayerController, SimStateClientClockPublisher);
```
Tadinu: commented codes without annotation would trigger certain wonders to future dev thus should be kept in #if RAPYUTA_SIM_DEBUG with notes or removed

Source/RapyutaSimulationPlugins/Private/Core/RRNetworkPlayerController.cpp
```diff
@@ -103,184 +110,49 @@ void ARRNetworkPlayerController::InitSimStateClientROS2()
...
-    }
+    UE_LOG(LogRapyutaCore,
+                Warning,
+                TEXT("[%s] [ARRNetworkPlayerController::BeginPlay]"), *GetName());
 
     if (IsLocalController())
     {
         FTimerManager& timerManager = GetWorld()->GetTimerManager();
-        timerManager.SetTimer(PossessTimerHandle, this, &ARRNetworkPlayerController::WaitToPossessPawn, 1.f, true);
+        // timerManager.SetTimer(PossessTimerHandle, this, &ARRNetworkPlayerController::WaitToPossessPawn, 1.f, true);
```
Tadinu: cmted code

Source/RapyutaSimulationPlugins/Public/Tools/ROS2Spawnable.h
```diff
@@ -51,11 +51,21 @@ class RAPYUTASIMULATIONPLUGINS_API UROS2Spawnable : public UActorComponent
...
+
+    UFUNCTION(BlueprintCallable)
+    virtual void SetNetworkPlayerId(const int32 InNetworkPlayerId);
 
 protected:
     virtual void OnComponentCreated() override;
+
+    //! Player ID No. This is used to check which network client posses this Robot.
+    UPROPERTY(Replicated)
+    int32 NetworkPlayerId;
```
Tadinu, line 70: `int32 NetworkPlayerId = 0;` or -1?

Source/RapyutaSimulationPlugins/Public/Core/RRNetworkPlayerController.h
```diff
@@ -31,70 +32,46 @@ class RAPYUTASIMULATIONPLUGINS_API ARRNetworkPlayerController : public APlayerCo
...
-    //! Client - Init possessed robot's ROS2Interface & MoveComponent, etc.
     UFUNCTION(Client, Reliable)
-    void ClientInitPawn(AActor* InActor);
-
-    // PLAYER
-    //! Controller's PlayerName, taken from [robotname] param in Engine build.
-    //! In Editor build, it will be automatically set from #PossessedPawn's name upon possession
-    UPROPERTY(EditAnywhere, BlueprintReadWrite, meta = (ExposeOnSpawn = "true"), Replicated)
-    FString PlayerName;
+    void InitSimStateClientROS2();
```
Tadinu:
```suggestion
    void ClientInitSimStateClientROS2();
```
with `Client/Server` prefix as a [convention](https://docs.unrealengine.com/4.26/en-US/InteractiveExperiences/Networking/Actors/RPCs/#quicktips) for an RPC method

Source/RapyutaSimulationPlugins/Public/Core/RRNetworkPlayerController.h
```diff
@@ -31,70 +32,46 @@ class RAPYUTASIMULATIONPLUGINS_API ARRNetworkPlayerController : public APlayerCo
...
+    FString PlayerName;
 
     /**
-     * @brief Wait to possess the targeted pawn
+     * @brief Create ROS2SimStateClient without initializing yet
      */
-    UFUNCTION()
-    void WaitToPossessPawn();
+    UFUNCTION(BlueprintCallable)
+    void CreateROS2SimStateClient(const TSubclassOf<URRROS2SimulationStateClient>& InSimStateClientClass);
```
Tadinu, line 75: Since it seems we init `ROS2SimStateClient` for each client separately below, should this be also with `UFUNCTION(Client, Reliable)`?
```suggestion
    void ClientCreateROS2SimStateClient(const TSubclassOf<URRROS2SimulationStateClient>& InSimStateClientClass);
```

Source/RapyutaSimulationPlugins/Private/Tools/SimulationState.cpp
```diff
@@ -309,13 +310,15 @@ AActor* ASimulationState::ServerSpawnEntity(const FROSSpawnEntityRequest& InROSS
...
+    
     // Wrap [InROSSpawnRequest] into a ROS2 spawnable component
     UROS2Spawnable* spawnableComponent = NewObject<UROS2Spawnable>(newEntity, TEXT("ROS2 Spawn Parameters"));
+    spawnableComponent->SetIsReplicated(true);
     spawnableComponent->RegisterComponent();
+    spawnableComponent->SetNetworkPlayerId(InNetworkPlayerId);
     spawnableComponent->InitializeParameters(InROSSpawnRequest);
 
-    newEntity->AddInstanceComponent(spawnableComponent);
+    // newEntity->AddInstanceComponent(spawnableComponent);
```
Tadinu: cmted code

Source/RapyutaSimulationPlugins/Private/Tools/RRROS2ClockPublisher.cpp
```diff
@@ -16,6 +16,8 @@ void URRROS2ClockPublisher::InitializeWithROS2(AROS2Node* InROS2Node)
 
 void URRROS2ClockPublisher::UpdateMessage(UROS2GenericMsg* InMessage)
 {
-    // Noted: Elapsed time: time in seconds since world was brought up for play
-    CastChecked<UROS2ClockMsg>(InMessage)->Update(GetWorld()->GetGameState()->GetServerWorldTimeSeconds());
+    // Noted: Elapsed time: time in seconds since world was brought up for play    
+    if (GetWorld()->GetGameState() != nullptr) {
+        CastChecked<UROS2ClockMsg>(InMessage)->Update(GetWorld()->GetGameState()->GetServerWorldTimeSeconds());
+    }
```
Tadinu:
```suggestion
    auto* gameState = GetWorld()->GetGameState();
    if (gameState)
    {
        CastChecked<UROS2ClockMsg>(InMessage)->Update(gameState->GetServerWorldTimeSeconds());
    }
```

Source/RapyutaSimulationPlugins/Private/Robots/RRBaseRobot.cpp
```diff
@@ -121,8 +265,21 @@ void ARRBaseRobot::GetLifetimeReplicatedProps(TArray<FLifetimeProperty>& OutLife
...
+    DOREPLIFETIME(ARRBaseRobot, ROSSpawnParameters);
+    DOREPLIFETIME(ARRBaseRobot, StartStopROS2Interface);
+}
+
+bool ARRBaseRobot::ReplicateSubobjects(UActorChannel *Channel, FOutBunch *Bunch, FReplicationFlags *RepFlags) 
+{
+    bool bWroteSomething = Super::ReplicateSubobjects(Channel, Bunch, RepFlags);
+    
+    // Single Object
+    bWroteSomething |= Channel->ReplicateSubobject(ROS2Interface, *Bunch, *RepFlags);
```
Tadinu, line 269: Oh, so if this is the way to allow us replicate a subobject, maybe we could turn `URRROS2SimulationStateClient` to an `UObject` & add it here too

Source/RapyutaSimulationPlugins/Private/Robots/RRRobotBaseVehicle.cpp
```diff
@@ -103,31 +102,47 @@ void ARRRobotBaseVehicle::SetAngularVel(const FVector& InAngularVel)
     ClientSetAngularVel(InAngularVel);
 }
 
-void ARRRobotBaseVehicle::ServerSetLinearVel_Implementation(float InClientTimeStamp,
+void ARRRobotBaseVehicle::ServerSetLinearVel(float InClientTimeStamp,
                                                             const FVector& InClientRobotPosition,
+                                                            const FQuat& InClientRobotQuat,
```
Tadinu:
```suggestion
void ARRRobotBaseVehicle::ServerSetLinearVel(float InClientTimeStamp,
                                                            const FTransform& InClientRobotTransform,
```

Source/RapyutaSimulationPlugins/Public/Robots/RRRobotBaseVehicle.h
```diff
@@ -115,29 +115,33 @@ class RAPYUTASIMULATIONPLUGINS_API ARRRobotBaseVehicle : public ARRBaseRobot
      * @param InClientTimeStamp
      * @param InClientRobotPosition
      * @param InLinearVel
+     * @note can't use rpc since this is not controlled by Paleyr. should add (Server, Reliable)
      */
-    UFUNCTION(BlueprintCallable, Server, Reliable)
-    virtual void ServerSetLinearVel(float InClientTimeStamp, const FVector& InClientRobotPosition, const FVector& InLinearVel);
+    UFUNCTION(BlueprintCallable)
+    virtual void ServerSetLinearVel(float InClientTimeStamp, const FVector& InClientRobotPosition, const FQuat& InClientRobotQuat, const FVector& InLinearVel);
```
Tadinu: If this is not an RPC anymore, we could remove `Server` prefix?

Source/RapyutaSimulationPlugins/Public/Robots/RRRobotBaseVehicle.h
```diff
@@ -115,29 +115,33 @@ class RAPYUTASIMULATIONPLUGINS_API ARRRobotBaseVehicle : public ARRBaseRobot
...
     /**
      * @brief Set server angular velocity to #RobotVehicleMoveComponent
      * @param InClientTimeStamp
      * @param InClientRobotRotation
      * @param InAngularVel
+     * @note can't use rpc since this is not controlled by Paleyr. should add (Server, Reliable)
      */
-    UFUNCTION(BlueprintCallable, Server, Reliable)
+    UFUNCTION(BlueprintCallable)
     virtual void ServerSetAngularVel(float InClientTimeStamp, const FRotator& InClientRobotRotation, const FVector& InAngularVel);
```
Tadinu: Same as above if not an RPC

Source/RapyutaSimulationPlugins/Public/Robots/RRRobotBaseVehicle.h
```diff
@@ -115,29 +115,33 @@ class RAPYUTASIMULATIONPLUGINS_API ARRRobotBaseVehicle : public ARRBaseRobot
...
 
     /**
      * @brief Set client linear velocity to #RobotVehicleMoveComponent
+     * @note can't use rpc since this is not controlled by Paleyr. should add (Client, Reliable)
      */
-    UFUNCTION(BlueprintCallable, Client, Reliable)
+    UFUNCTION(BlueprintCallable)
     virtual void ClientSetLinearVel(const FVector& InLinearVel);
```
Tadinu: Same as above, `Client` prefix to be removed

Source/RapyutaSimulationPlugins/Public/Robots/RRRobotBaseVehicle.h
```diff
@@ -115,29 +115,33 @@ class RAPYUTASIMULATIONPLUGINS_API ARRRobotBaseVehicle : public ARRBaseRobot
...
 
     /**
      * @brief Set server angular velocity to #RobotVehicleMoveComponent
+     * @note can't use rpc since this is not controlled by Paleyr. should add (Client, Reliable)
      */
-    UFUNCTION(BlueprintCallable, Client, Reliable)
+    UFUNCTION(BlueprintCallable)
     virtual void ClientSetAngularVel(const FVector& InAngularVel);
```
Tadinu: Same as above for non-RPC

Source/RapyutaSimulationPlugins/Public/Tools/RRROS2SimulationStateClient.h
```diff
@@ -162,9 +162,33 @@ class RAPYUTASIMULATIONPLUGINS_API URRROS2SimulationStateClient : public UActorC
...
+    {
+        return NetworkPlayerId;
+    }
+
 protected:
     virtual void OnComponentCreated() override;
 
+    //! NetworkPlayerId which is used to differenciate client in server.
+    UPROPERTY(BlueprintReadOnly, Replicated)
+    int32 NetworkPlayerId;
```
Tadinu, line 190: `int32 NetworkPlayerId = 0;` or -1?

Source/RapyutaSimulationPlugins/Private/Tools/SimulationState.cpp
```diff
@@ -309,13 +310,15 @@ AActor* ASimulationState::ServerSpawnEntity(const FROSSpawnEntityRequest& InROSS
     {
         return nullptr;
     }
-
+    
     // Wrap [InROSSpawnRequest] into a ROS2 spawnable component
     UROS2Spawnable* spawnableComponent = NewObject<UROS2Spawnable>(newEntity, TEXT("ROS2 Spawn Parameters"));
+    spawnableComponent->SetIsReplicated(true);
```
Tadinu: This is already added in `UROS2Spawnable::OnComponentCreated()` due to a suggestive warning on `SetIsReplicated()` should be called during component [construction](https://docs.unrealengine.com/4.27/en-US/InteractiveExperiences/Networking/Actors/Components/#:~:text=done%20in%20the-,class%20constructor,-after%20spawning%20the).

Side notes for further refactoring, similar to ROS2Interface, this could be an `UObject`, then get to be:
* Created inside `ARRBaseRobot`
* Added to `::ReplicateSubobjects()`
* Fetched directly instead of using the expensive `FindComponentByClass`

Source/RapyutaSimulationPlugins/Public/Robots/RRBaseRobot.h
```diff
@@ -55,14 +71,59 @@ class RAPYUTASIMULATIONPLUGINS_API ARRBaseRobot : public ARRBaseActor
...
+
+    UFUNCTION(BlueprintCallable)
+    virtual void OnRep_StartStopROS2Interface();
+
+    /**
+     * @brief Check necessary variables has initialized and PlayerId which spaned robot is match the this client PlayerId
+     * @return true if playerId matches robot spawn playerId
+     */
+    bool IsAutorizedInThisClient();
```
Tadinu:
```suggestion
    bool IsAuthorizedInThisClient();
```

Source/RapyutaSimulationPlugins/Private/Tools/SimulationState.cpp
```diff
@@ -324,11 +327,16 @@ AActor* ASimulationState::ServerSpawnEntity(const FROSSpawnEntityRequest& InROSS
     if (robot)
     {
         robot->RobotUniqueName = InROSSpawnRequest.StateName;
+
+        // todo child actor component is not replicated.
+        // https://forums.unrealengine.com/t/child-actor-component-never-replicated/23497/18
+        robot->ROSSpawnParameters = spawnableComponent;
+        robot->ServerRobot = robot;
```
Tadinu, line 333: * `ROSSpawnParameters` is a dynamic component, so should get [replicated](https://docs.unrealengine.com/4.27/en-US/InteractiveExperiences/Networking/Actors/Components/#:~:text=server%20and%20client.-,Dynamic%20components,-are%20components%20spawned) with `SetIsReplicated(true)`?
* Just to be curious, instead of storing `ServerRobot` for its client version, in any case could an RPC method be used for invoking that ServerRobot's function?

Source/RapyutaSimulationPlugins/Private/Tools/SimulationState.cpp
```diff
@@ -346,7 +354,7 @@ AActor* ASimulationState::ServerSpawnEntity(const FROSSpawnEntityRequest& InROSS
     return newEntity;
 }
 
-AActor* ASimulationState::ServerSpawnEntity(const FROSSpawnEntityRequest& InRequest)
+AActor* ASimulationState::ServerSpawnEntity(const FROSSpawnEntityRequest& InRequest, const int32& InNetworkPlayerId)
```
Tadinu:
```suggestion
AActor* ASimulationState::ServerSpawnEntity(const FROSSpawnEntityRequest& InRequest, const int32 InNetworkPlayerId)
```
as POD, if passed as const, should be passed by value in general
https://www.sandordargo.com/blog/2020/11/25/when-use-const-4-parameters
https://www.faqcode4u.com/faq/476434/does-const-call-by-reference-improve-performance-when-applied-to-primitive-types

Source/RapyutaSimulationPlugins/Public/Core/RRNetworkPlayerController.h
```diff
@@ -137,8 +114,31 @@ class RAPYUTASIMULATIONPLUGINS_API ARRNetworkPlayerController : public APlayerCo
...
+     * @param InClientRobotPosition
+     * @param InLinearVel
+     * @note can't use rpc since this is not controlled by Paleyr. should add (Server, Reliable)
+     */
+    UFUNCTION(BlueprintCallable, Server, Reliable)
+    virtual void ServerSetLinearVel(ARRBaseRobot* InServerRobot,float InClientTimeStamp, const FVector& InClientRobotPosition, const FQuat& InClientRobotQuat, const FVector& InLinearVel);
+
+    /**
+     * @brief Set server angular velocity to #RobotVehicleMoveComponent. Non player object can use this method to call RPC.
+     * @param InServerRobot target robot
```
Tadinu:
```suggestion
     * @param InServerRobot target server-owned robot
```

Source/RapyutaSimulationPlugins/Public/Core/RRNetworkPlayerController.h
```diff
@@ -137,8 +114,31 @@ class RAPYUTASIMULATIONPLUGINS_API ARRNetworkPlayerController : public APlayerCo
      */
     void UpdateLocalClock(float InDeltaSeconds);
 
+
+    /**
+     * @brief Set server linear velocity to #RobotVehicleMoveComponent. Non player object can use this method to call RPC.
+     * @param InServerRobot target robot
```
Tadinu:
```suggestion
     * @param InServerRobot target server-owned robot
```

Source/RapyutaSimulationPlugins/Public/Robots/RRBaseRobot.h
```diff
@@ -55,14 +71,59 @@ class RAPYUTASIMULATIONPLUGINS_API ARRBaseRobot : public ARRBaseActor
...
     /**
      * @brief Instantiate ROS2 Interface without initializing yet
+     * @note can't use rpc since this is not controlled by Paleyr. should add (Server, Reliable)
      */
+    UFUNCTION(BlueprintCallable)
     void CreateROS2Interface();
 
+    /**
+     * @brief Initialize ROS2 Interface
+     * @note can't use rpc since this is not controlled by Paleyr. should add (Client, Reliable)
```
Tadinu:
```suggestion
     * @note This is meant to be called in server only, but can't be an RPC since the robot is not always owned by the same [connection](https://docs.unrealengine.com/4.27/en-US/InteractiveExperiences/Networking/Actors/OwningConnections) with the client's PlayerController.
```

Source/RapyutaSimulationPlugins/Public/Robots/RRBaseRobot.h
```diff
@@ -55,14 +71,59 @@ class RAPYUTASIMULATIONPLUGINS_API ARRBaseRobot : public ARRBaseActor
...
+     * @return true if playerId matches robot spawn playerId
+     */
+    bool IsAutorizedInThisClient();
+
+    //!ROSSpawn parameters which is passed to ROS2Interface
+    UPROPERTY(VisibleAnywhere, Replicated)
+    UROS2Spawnable* ROSSpawnParameters = nullptr;
+
+    /**
+     * @brief Pointer to the robot in the server
```
Tadinu:
```suggestion
     * @brief Pointer to the robot's server-owned version
```

Source/RapyutaSimulationPlugins/Public/Robots/RRBaseRobot.h
```diff
@@ -55,14 +71,59 @@ class RAPYUTASIMULATIONPLUGINS_API ARRBaseRobot : public ARRBaseActor
...
+    /**
+     * @brief Initialize ROS2 Interface
+     * @note can't use rpc since this is not controlled by Paleyr. should add (Client, Reliable)
+     */
+    UFUNCTION(BlueprintCallable)
+    void InitROS2Interface();
+
+    /**
+     * @brief Stop ROS2 Interface
+     * @note can't use rpc since this is not controlled by Paleyr. should add (Client, Reliable)
```
Tadinu: same note as above

Source/RapyutaSimulationPlugins/Private/Robots/RRRobotVehicleROSController.cpp
```diff
@@ -10,19 +10,19 @@ void ARRRobotVehicleROSController::OnPossess(APawn* InPawn)
...
 }
 
 void ARRRobotVehicleROSController::OnUnPossess()
 {
-    if (ROS2Interface)
+    auto* robotVehicle = GetPawn<ARRBaseRobot>();
+    if (robotVehicle)
     {
-        ROS2Interface->StopPublishers();
+        robotVehicle->StopROS2Interface();
```
Tadinu:
```
auto* robot = CastChecked<ARRBaseRobot>(GetPawn());
robot->StopROS2Interface();
```

Source/RapyutaSimulationPlugins/Private/Robots/RRRobotVehicleROSController.cpp
```diff
@@ -10,19 +10,19 @@ void ARRRobotVehicleROSController::OnPossess(APawn* InPawn)
...
-    // Refer to ARRBaseRobot::CreateROS2Interface() for reasons why it is inited here but not earlier
-    auto* robotVehicle = CastChecked<ARRRobotBaseVehicle>(InPawn);
-    verify(IsValid(robotVehicle->ROS2Interface));
-    ROS2Interface = robotVehicle->ROS2Interface;
-    ROS2Interface->Initialize(robotVehicle);
+    auto* robotVehicle = GetPawn<ARRRobotBaseVehicle>();
+    if(robotVehicle)
+    {
+        robotVehicle->InitROS2Interface();
+    }
```
Tadinu:
```suggestion
    auto* robot = CastChecked<ARRBaseRobot>(GetPawn());
    robot->InitROS2Interface();
```
For another PR: this class could be renamed to `ARRRobotROSController` since it seems to have no vehicle-related functions

Tadinu (commented):
Thanks for the cool PR.
Some clarification maybe, if I understand correctly, should be added to the description regarding robot possession & its owning connection:
* Since a NetworkPlayerController could only possess a single robot as its PlayerPawn at a time, for multiple robots in one client, each of them now is assigned a `NetworkPlayerId` (through its `ROS2SpawnParameters`) to match the NetworkPlayerController's PlayerId to call `InitROS2Interface()`
* Robot possession is then reverted to be done by its AIController, not NetworkPlayerController
Detailed highlights:
* `ARRBaseRobot`
   * Add `IsAuthorizedInThisClient()`
   * Add `ServerRobot` as its server-owned version
   * Add `ROS2SpawnParameters` with `NetworkPlayerId` being passed at spawning
   * Certain members functions `CreateROS2Interface()/InitROS2Interface()` are non-RPC, triggered by added `bStartStopROS2Interface`
   * Add `ERRNetworkAuthorityType NetworkAuthorityType`
* `ARRRobotBaseVehicle`
   * `SetLinearVel()/SetAngularVel()` reverted to non-RPC
* `ARRNetworkPlayerController`
   * Add RPC `ServerSetLinearVel()/ServerSetAngularVel()`
* `URRRobotROS2Interface` Add `ROSSpawnParameters`, pointing to `ARRBaseRobot::ROSSpawnParameters` which is created at its spawning, to fetch its ros namespace directly for its `RobotROS2Node`, without using `FindComponentByClass`

Source/RapyutaSimulationPlugins/Private/Core/RRCrowdROSController.cpp
```diff
@@ -15,17 +15,19 @@ void ARRCrowdROSController::OnPossess(APawn* InPawn)
...
     //  may have not been fully accessible until now
-    auto* robot = CastChecked<ARRBaseRobot>(InPawn);
-    verify(IsValid(robot->ROS2Interface));
-    ROS2Interface = robot->ROS2Interface;
-    ROS2Interface->Initialize(robot);
+    auto* robotVehicle = GetPawn<ARRBaseRobot>();
+    if(robotVehicle)
+    {
+        robotVehicle->InitROS2Interface();
+    }
```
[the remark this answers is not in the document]
Tadinu: `robotVehicle` -> `robot` to be agnostic
```
auto* robot = CastChecked<ARRBaseRobot>(GetPawn());
robot->InitROS2Interface();
```
as similar to above

Source/RapyutaSimulationPlugins/Private/Core/RRCrowdROSController.cpp
```diff
@@ -15,17 +15,19 @@ void ARRCrowdROSController::OnPossess(APawn* InPawn)
...
 }
 
 void ARRCrowdROSController::OnUnPossess()
 {
-    if (ROS2Interface)
+    auto* robotVehicle = GetPawn<ARRBaseRobot>();
+    if (robotVehicle)
     {
-        ROS2Interface->StopPublishers();
+        robotVehicle->StopROS2Interface();
```
Tadinu:
```suggestion
    auto* robot = CastChecked<ARRBaseRobot>(InPawn);
    robot->StopROS2Interface();
```
* Here we cast APawn -> ARRBaseRobot, so the robot var may not necessarily be a vehicle
* `GetPawn<>()` is not as overt as `InPawn`, though it is assumed to have already been set as AIController::Pawn beforehands, & it does not use CastChecked, which is stricter due to including null-check & thus intentional here since all of our robot classes should essentially be an `ARRBaseRobot`

Source/RapyutaSimulationPlugins/Private/Core/RRNetworkPlayerController.cpp
```diff
@@ -34,28 +33,21 @@ ARRNetworkPlayerController::ARRNetworkPlayerController()
...
+    // DOREPLIFETIME(ARRNetworkPlayerController, SimStateClientROS2Node);
+    // DOREPLIFETIME(ARRNetworkPlayerController, SimStateClientClockPublisher);
     DOREPLIFETIME(ARRNetworkPlayerController, ServerSimState);
     DOREPLIFETIME(ARRNetworkPlayerController, ROS2SimStateClient);
     DOREPLIFETIME(ARRNetworkPlayerController, PlayerName);
 
     // After pawn possession, PossessedPawn & Super::Pawn are actually referring to the same one.
     // But we could not setup replication for Pawn here due to its being private in Super
-    DOREPLIFETIME(ARRNetworkPlayerController, PossessedPawn);
+    // DOREPLIFETIME(ARRNetworkPlayerController, PossessedPawn);
```
Tadinu: cmted code -> removed?

yuokamoto:
> Thanks for the cool PR. Some clarification maybe, if I understand correctly, should be added to the description regarding robot possession & its owning connection:
> 
> * Since a NetworkPlayerController could only possess a single robot as its PlayerPawn at a time, for multiple robots in one client, each of them now is assigned a `NetworkPlayerId` (through its `ROS2SpawnParameters`) to match the NetworkPlayerController's PlayerId to call `InitROS2Interface()`
> * Robot possession is then reverted to be done by its AIController, not NetworkPlayerController
>   Detailed highlights:
> * `ARRBaseRobot`
>   
>   * Add `IsAuthorizedInThisClient()`
>   * Add `ServerRobot` as its server-owned version
>   * Add `ROS2SpawnParameters` with `NetworkPlayerId` being passed at spawning
>   * Certain members functions `CreateROS2Interface()/InitROS2Interface()` are non-RPC, triggered by added `bStartStopROS2Interface`
>   * Add `ERRNetworkAuthorityType NetworkAuthorityType`
> * `ARRRobotBaseVehicle`
>   
>   * `SetLinearVel()/SetAngularVel()` reverted to non-RPC
> * `ARRNetworkPlayerController`
>   
>   * Add RPC `ServerSetLinearVel()/ServerSetAngularVel()`
> * `URRRobotROS2Interface` Add `ROSSpawnParameters`, pointing to `ARRBaseRobot::ROSSpawnParameters` which is created at its spawning, to fetch its ros namespace directly for its `RobotROS2Node`, without using `FindComponentByClass`

yes. that is correct.

ExternalTest/rr_sim_tests/rr_sim_tests/utils/utils.py
yuokamoto: it don't work I think but since we don't support windows yet, it can be ignoreble

Source/RapyutaSimulationPlugins/Private/Core/RRCrowdROSController.cpp
yuokamoto: CastChecked stop the simulator if it is nullptr. it is better to avoid crashing editor/game if that is nullptr.

Source/RapyutaSimulationPlugins/Public/Tools/ROS2Spawnable.h
yuokamoto: -1 maybe good but need to check to avoid intiializing with id which can be generated by UE.

Source/RapyutaSimulationPlugins/Public/Core/RRNetworkPlayerController.h
yuokamoto: ROS2SimStateClient need to be in the client as well to use RPC.

Source/RapyutaSimulationPlugins/Private/Tools/SimulationState.cpp
yuokamoto: - I think it is set in constructor already
- I tried but I can't find option. got this error "No owning connection for actor *. Function * will not be processed”"

Source/RapyutaSimulationPlugins/Private/Core/RRCrowdROSController.cpp
Tadinu: Yeah I see, generally an exception (user-oriented) should not be mistaken as a 
design error (dev-oriented).
* We should definitely use `Cast` only for exception but `CastChecked` for the latter like an assert to help detect early error in the design.
* In this specific case, it must **never** be a nullptr essentially since this class (as having its ROS in its name) is meant to work with out ROS2Interface, which requires itself to inherit from `ARRBaseRobot`

Source/RapyutaSimulationPlugins/Private/Robots/RRBaseRobot.cpp
yuokamoto: UObject can't use RPC

Source/RapyutaSimulationPlugins/Private/Core/RRCrowdROSController.cpp
yuokamoto: Controller pauses/unpauses can be happen from blueprint/c++ dynamically as well. Since restarting edtior would slow down develpment, I prefer not to exit editor.

ExternalTest/rr_sim_tests/rr_sim_tests/test_random_spawn.py
yuokamoto: `it seems meant to only break out of while loop, so could it be while True & break?` I don't get it

Source/RapyutaSimulationPlugins/Private/Core/RRCrowdROSController.cpp
Tadinu: Ah I see, so in that case InPawn is null by itself

ExternalTest/rr_sim_tests/rr_sim_tests/test_random_spawn.py
Tadinu: Sorry, here you got nested loops, so just to make it clear, could `res` be renamed as like `not_yet_spawned`?

ExternalTest/rr_sim_tests/rr_sim_tests/utils/utils.py
Tadinu: Yeah I mean if ROS' `/` does not realy equate the path delimiter, using `os.path.join` might be a bit misleading & also a bit of overkill

Source/RapyutaSimulationPlugins/Public/Robots/RRBaseRobot.h
Tadinu: Seems not yet updated?

Source/RapyutaSimulationPlugins/Private/Core/RRCrowdROSController.cpp
Tadinu: Also `robotVehicle` -> `robot`?

Tadinu: Ah I see, so in those cases, InPawn is null by itself, so a log is good.
Besides robotVehicle -> robot?

Source/RapyutaSimulationPlugins/Private/Core/RRCrowdROSController.cpp
Tadinu: robotVehicle -> robot

Tadinu:
Thanks for addressing my comments. Just a small note on `can't use rpc since this is not controlled by Paleyr` which seems to linger in various places, thus might need to be updated like the comment for `ARRBaseRobot::InitROS2Interface()`?

Source/RapyutaSimulationPlugins/Private/Core/RRNetworkPlayerController.cpp
```diff
@@ -103,184 +103,48 @@ void ARRNetworkPlayerController::InitSimStateClientROS2()
     // ClockPublisher's RegisterComponent() is done by [AROS2Node::AddPublisher()]
     SimStateClientClockPublisher->InitializeWithROS2(SimStateClientROS2Node);
 
-    UE_LOG(LogRapyutaCore, Warning, TEXT("[%s] SimStateClient ROS2Node[%s] created"), *PlayerName, *SimStateClientROS2Node->Name);
+    UE_LOG(LogRapyutaCore, Error, TEXT("[%s] SimStateClient ROS2Node[%s] created"), *PlayerName, *SimStateClientROS2Node->Name);
```
Tadinu:
```suggestion
    UE_LOG(LogRapyutaCore, Log, TEXT("[%s] SimStateClient ROS2Node[%s] created"), *PlayerName, *SimStateClientROS2Node->Name);
```

yuokamoto:
ci passed in turtlebot3-UE
https://github.com/rapyuta-robotics/turtlebot3-UE/pull/38

Source/RapyutaSimulationPlugins/Private/Core/RRNetworkGameMode.cpp
```diff
@@ -5,12 +5,14 @@
...
 #include "Core/RRNetworkGameState.h"
 #include "Core/RRNetworkPlayerController.h"
+#include "Tools/GhostPlayerPawn.h"
 #include "Tools/RRROS2SimulationStateClient.h"
 
 ARRNetworkGameMode::ARRNetworkGameMode()
 {
     GameStateClass = ARRNetworkGameState::StaticClass();
     PlayerControllerClass = ARRNetworkPlayerController::StaticClass();
+    DefaultPawnClass = AGhostPlayerPawn::StaticClass();
```
Tadinu: Could we try `DefaultPawnClass = nullptr;`? since `AGhostPlayerPawn` seems to have no particular function

Source/RapyutaSimulationPlugins/Public/Core/RRNetworkPlayerController.h
```diff
@@ -137,8 +121,35 @@ class RAPYUTASIMULATIONPLUGINS_API ARRNetworkPlayerController : public APlayerCo
      */
     void UpdateLocalClock(float InDeltaSeconds);
 
+    /**
+     * @brief Set server linear velocity to #RobotVehicleMoveComponent. Non player object can use this method to call RPC.
+     * @param InServerRobot target server-owned robot
+     * @param InClientTimeStamp
+     * @param InClientRobotPosition
+     * @param InLinearVel
+     * @note can't use rpc since this is not controlled by Paleyr. should add (Server, Reliable)
```
Tadinu:
```suggestion
```
This seems to be an RPC now already?

Source/RapyutaSimulationPlugins/Public/Core/RRNetworkPlayerController.h
```diff
@@ -137,8 +121,35 @@ class RAPYUTASIMULATIONPLUGINS_API ARRNetworkPlayerController : public APlayerCo
...
+                                    const FTransform& InClientRobotTransform,
+                                    const FVector& InLinearVel);
+
+    /**
+     * @brief Set server angular velocity to #RobotVehicleMoveComponent. Non player object can use this method to call RPC.
+     * @param InServerRobot target server-owned robot
+     * @param InClientTimeStamp
+     * @param InClientRobotRotation
+     * @param InAngularVel
+     * @note can't use rpc since this is not controlled by Paleyr. should add (Server, Reliable)
```
Tadinu:
```suggestion
```
This seems to be an RPC now with (Server, Reliable)?

Source/RapyutaSimulationPlugins/Public/Robots/RRBaseRobot.h
```diff
@@ -55,14 +70,65 @@ class RAPYUTASIMULATIONPLUGINS_API ARRBaseRobot : public ARRBaseActor
...
+    /**
+     * @brief Pointer to the robot's server-owned version
+     * @note Owner can't be used since non-player pawn don't have that.
+     */
+    UPROPERTY(VisibleAnywhere, Replicated)
+    ARRBaseRobot* ServerRobot = nullptr;
+
     /**
      * @brief Instantiate ROS2 Interface without initializing yet
+     * @note can't use rpc since this is not controlled by Paleyr. should add (Server, Reliable)
```
Tadinu:
```suggestion
@note Not uses RPC since the robot is not always owned by the same
     * [connection](https://docs.unrealengine.com/4.27/en-US/InteractiveExperiences/Networking/Actors/OwningConnections) with the
     * client's PlayerController.
```

Source/RapyutaSimulationPlugins/Public/Robots/RRRobotBaseVehicle.h
```diff
@@ -111,34 +117,42 @@ class RAPYUTASIMULATIONPLUGINS_API ARRRobotBaseVehicle : public ARRBaseRobot
     virtual void SetAngularVel(const FVector& InAngularVel);
 
     /**
-     * @brief Set server linear velocity to #RobotVehicleMoveComponent
+     * @brief Set server position and linear velocity to #RobotVehicleMoveComponent
      * @param InClientTimeStamp
      * @param InClientRobotPosition
      * @param InLinearVel
+     * @note can't use rpc since this is not controlled by Paleyr. should add (Server, Reliable)
```
Tadinu:
```suggestion
@note Not uses RPC since the robot is not always owned by the same
     * [connection](https://docs.unrealengine.com/4.27/en-US/InteractiveExperiences/Networking/Actors/OwningConnections) with the
     * client's PlayerController.
```

Source/RapyutaSimulationPlugins/Public/Robots/RRRobotBaseVehicle.h
```diff
@@ -111,34 +117,42 @@ class RAPYUTASIMULATIONPLUGINS_API ARRRobotBaseVehicle : public ARRBaseRobot
...
+                                          const FTransform& InClientRobotTransform,
+                                          const FVector& InLinearVel);
 
     /**
-     * @brief Set server angular velocity to #RobotVehicleMoveComponent
+     * @brief Set server rotation and angular velocity to #RobotVehicleMoveComponent
      * @param InClientTimeStamp
      * @param InClientRobotRotation
      * @param InAngularVel
+     * @note can't use rpc since this is not controlled by Paleyr. should add (Server, Reliable)
```
Tadinu:
```suggestion
@note Not uses RPC since the robot is not always owned by the same
     * [connection](https://docs.unrealengine.com/4.27/en-US/InteractiveExperiences/Networking/Actors/OwningConnections) with the
     * client's PlayerController.
```

Source/RapyutaSimulationPlugins/Public/Robots/RRRobotBaseVehicle.h
```diff
@@ -111,34 +117,42 @@ class RAPYUTASIMULATIONPLUGINS_API ARRRobotBaseVehicle : public ARRBaseRobot
...
+     * @note can't use rpc since this is not controlled by Paleyr. should add (Client, Reliable)
      */
-    UFUNCTION(BlueprintCallable, Client, Reliable)
-    virtual void ClientSetLinearVel(const FVector& InLinearVel);
+    UFUNCTION(BlueprintCallable)
+    virtual void SetLocalLinearVel(const FVector& InLinearVel);
 
     /**
      * @brief Set server angular velocity to #RobotVehicleMoveComponent
+     * @note can't use rpc since this is not controlled by Paleyr. should add (Client, Reliable)
```
Tadinu:
```suggestion
     * @brief Set local angular velocity to #RobotVehicleMoveComponent
     * @note Not uses RPC since the robot is not always owned by the same
     * [connection](https://docs.unrealengine.com/4.27/en-US/InteractiveExperiences/Networking/Actors/OwningConnections) with the
     * client's PlayerController.
```

Source/RapyutaSimulationPlugins/Public/Robots/RRRobotBaseVehicle.h
```diff
@@ -111,34 +117,42 @@ class RAPYUTASIMULATIONPLUGINS_API ARRRobotBaseVehicle : public ARRBaseRobot
...
-    UFUNCTION(BlueprintCallable, Server, Reliable)
-    virtual void ServerSetAngularVel(float InClientTimeStamp, const FRotator& InClientRobotRotation, const FVector& InAngularVel);
+    UFUNCTION(BlueprintCallable)
+    virtual void SyncServerAngularMovement(float InClientTimeStamp,
+                                           const FRotator& InClientRobotRotation,
+                                           const FVector& InAngularVel);
 
     /**
      * @brief Set client linear velocity to #RobotVehicleMoveComponent
+     * @note can't use rpc since this is not controlled by Paleyr. should add (Client, Reliable)
```
Tadinu:
```suggestion
     * @brief Set local linear velocity to #RobotVehicleMoveComponent
     * @note Not uses RPC since the robot is not always owned by the same
     * [connection](https://docs.unrealengine.com/4.27/en-US/InteractiveExperiences/Networking/Actors/OwningConnections) with the
     * client's PlayerController.
```

Source/RapyutaSimulationPlugins/Public/Tools/GhostPlayerPawn.h
```diff
@@ -0,0 +1,25 @@
...
+
+#include "GhostPlayerPawn.generated.h"
+
+class USphereComponent;
+
+/**
+ * @brief Default player pawn without collision and mesh
+ */
+UCLASS(ClassGroup = (Custom), Blueprintable, meta = (BlueprintSpawnableComponent))
+class RAPYUTASIMULATIONPLUGINS_API AGhostPlayerPawn : public ADefaultPawn
```
Tadinu: If `DefaultPawnClass = nullptr` in `ARRNetworkGameMode` works, we'd not need this class I think.
Besides, `RR` prefix also?, other classes in `Tools` folder should be updated as well to have the prefix. This seems to have happened in some PRs before, the meaning is mostly just:
* To quickly differentiate itself from UE or 3rd-plugin classes
* To provide better src uniformity, security & copyright

Source/RapyutaSimulationPlugins/Public/Core/RRNetworkPlayerController.h
larryng-rapyuta: I assume its because the robot couldnt use an rpc?
```suggestion
     * @note This method is because RRRobotBaseVehicle can't use rpc since it is not controlled/possessed by the Player. 
```

larryng-rapyuta: I assume its because the robot couldnt use an rpc?
```suggestion
     * @note This method is because RRRobotBaseVehicle can't use rpc since it is not controlled/possessed by the Player. 
```

Source/RapyutaSimulationPlugins/Public/Core/RRNetworkPlayerController.h
larryng-rapyuta: I assume its because the robot couldnt use an rpc?
```suggestion
     * @note This method is because RRRobotBaseVehicle can't use rpc since it is not controlled/possessed by the Player. 
```

Source/RapyutaSimulationPlugins/Private/Core/RRNetworkPlayerController.cpp
```diff
@@ -328,3 +199,49 @@ void ARRNetworkPlayerController::ReceivedPlayer()
         ClientRequestLocalClockUpdate(LocalTime);
     }
 }
+
+void ARRNetworkPlayerController::ServerSetLinearVel_Implementation(ARRBaseRobot* InServerRobot,
```
larryng-rapyuta, line 203: Maybe we can create a replicated ActorComponent that does these Set Velocity actions and attach it to the player controller for the robots to be able to pass these RPCs to? Kind of like how SimulationStateClient is. It could help with separating out the functions a bit.

We could even have one per robot, in case differentiation between the robots in the future is needed but not too sure if that would be a good idea for now

Source/RapyutaSimulationPlugins/Public/Tools/GhostPlayerPawn.h
yuokamoto: - DefaultPawnClass = nullptr does not work. What I need is DefaultPawn without mesh and collision. DefaultPawn has basic functionalities to be controleld by PlayerController
- will rename.

Source/RapyutaSimulationPlugins/Private/Core/RRNetworkPlayerController.cpp
yuokamoto: sounds good. 
but it may require a bit more time. let me work on different pr.

yuokamoto: https://www.wrike.com/workspace.htm?acc=1794223#folder/858739231/list?filters=status%3Dactive&sidePanelItemId=955024545&sortOrder=5&spaceId=364052400&viewId=74889003

Tadinu (approved):

ExternalTest/rr_sim_tests/rr_sim_tests/utils/wait_for_spawned_entity.py
```diff
@@ -9,10 +10,10 @@
 
 SERVICE_NAME_GET_ENTITY_STATE = 'GetEntityState'
 
-def wait_for_spawned_entity(in_entity_name, in_timeout=10.0):
+def wait_for_spawned_entity(in_entity_name, service_namespace='', in_timeout=10.0):
```
Tadinu: Ah we need to update
* [test_robot_spawn.py](https://github.com/rapyuta-robotics/RapyutaSimulationPlugins/blob/yu_client_server_update/ExternalTest/rr_sim_tests/rr_sim_tests/test_robot_spawn.py#L98)
* [test_robot_remove.py](https://github.com/rapyuta-robotics/RapyutaSimulationPlugins/blob/yu_client_server_update/ExternalTest/rr_sim_tests/rr_sim_tests/test_robot_remove.py#L74)
* [test_robot_spawn.py](https://github.com/rapyuta-robotics/RapyutaSimulationPlugins/blob/yu_client_server_update/ExternalTest/rr_sim_tests/rr_sim_tests/test_robot_spawn.py#L98)
* [utils.py](https://github.com/rapyuta-robotics/RapyutaSimulationPlugins/blob/yu_client_server_update/ExternalTest/rr_sim_tests/rr_sim_tests/utils/utils.py#L105)
* RapyutaSimInternal's [test_world_spawn.py](https://github.com/rapyuta-robotics/RapyutaSimulationInternal/blob/devel/ExternalTest/rr_sim_internal_tests/rr_sim_internal_tests/test_world_spawn.py#L115)

yuokamoto: I change the arg order to  avoid modification

PR #88, opened 2022-09-03
Update by autogen

yuokamoto:
udpate msg types to use files which is  generated from latest [UE_tools](https://github.com/rapyuta-robotics/UE_tools/pull/7)
https://github.com/rapyuta-robotics/rclUE/pull/55

### DIFF
--- a/Source/RapyutaSimulationPlugins/Private/Robots/RRBaseRobot.cpp
+++ b/Source/RapyutaSimulationPlugins/Private/Robots/RRBaseRobot.cpp
@@ -109,10 +109,6 @@ bool ARRBaseRobot::IsAuthorizedInThisClient()
 #endif
         res = false;
     }
-    else if (IsNetMode(NM_Standalone))    // Standalone
-    {
-        res = true;
-    }
     else if (nullptr == ROS2Interface->ROSSpawnParameters)
     {
 #if RAPYUTA_SIM_DEBUG
@@ -188,7 +184,7 @@ void ARRBaseRobot::InitROS2Interface()
     UE_LOG(LogRapyutaCore, Warning, TEXT("[%s][ARRBaseRobot::InitROS2Interface] %d"), *GetName(), IsAuthorizedInThisClient());
 #endif
 
-    if ((IsNetMode(NM_Standalone) || IsNetMode(NM_Client)) && IsAuthorizedInThisClient())
+    if ((IsNetMode(NM_Standalone) && nullptr != ROS2Interface) || (IsNetMode(NM_Client) && IsAuthorizedInThisClient()))
     {
         ROS2Interface->Initialize(this);
         if (NetworkAuthorityType == ERRNetworkAuthorityType::CLIENT)
@@ -210,7 +206,7 @@ void ARRBaseRobot::DeInitROS2Interface()
     UE_LOG(LogRapyutaCore, Warning, TEXT("[%s][ARRBaseRobot::StopROS2Interface] %d"), *GetName(), IsAuthorizedInThisClient());
 #endif
 
-    if (IsNetMode(NM_Standalone) || (IsNetMode(NM_Client) && IsAuthorizedInThisClient()))
+    if ((IsNetMode(NM_Standalone) && nullptr != ROS2Interface) || (IsNetMode(NM_Client) && IsAuthorizedInThisClient()))
     {
         ROS2Interface->DeInitialize();
     }
--- a/Source/RapyutaSimulationPlugins/Private/Robots/RRBaseRobot.cpp
+++ b/Source/RapyutaSimulationPlugins/Private/Robots/RRBaseRobot.cpp
@@ -7,7 +7,6 @@
 #include "Net/UnrealNetwork.h"
 
 // rclUE
-#include "Msgs/ROS2TFMsg.h"
 #include "ROS2Node.h"
 
 // RapyutaSimulationPlugins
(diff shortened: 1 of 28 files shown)


Source/RapyutaSimulationPlugins/Private/Robots/RRBaseRobot.cpp
```diff
@@ -189,7 +188,7 @@ void ARRBaseRobot::InitROS2Interface()
     UE_LOG(LogRapyutaCore, Warning, TEXT("[%s][ARRBaseRobot::InitROS2Interface] %d"), *GetName(), IsAuthorizedInThisClient());
 #endif
 
-    if (IsNetMode(NM_Standalone) || (IsNetMode(NM_Client) && IsAuthorizedInThisClient()))
+    if ((IsNetMode(NM_Standalone) || IsNetMode(NM_Client)) && IsAuthorizedInThisClient())
```
Tadinu: `IsAuthorizedInThisClient()` is also required for `NM_Standalone`?

Source/RapyutaSimulationPlugins/Private/Tools/RRROS2OdomPublisher.cpp
```diff
@@ -52,21 +52,21 @@ bool URRROS2OdomPublisher::GetOdomData(FROSOdometry& OutOdomData) const
 {
     const URobotVehicleMovementComponent* moveComponent =
         RobotVehicle.IsValid() ? RobotVehicle.Get()->RobotVehicleMoveComponent : nullptr;
-    if (moveComponent && moveComponent->OdomData.IsValid())
+    if (moveComponent)
```
Tadinu, line 55: Just a note about `OdomData.IsValid()`, which [checks](https://github.com/rapyuta-robotics/rclUE/blob/devel/Source/rclUE/Public/Msgs/ROS2OdometryMsg.h#L57) `PoseCovariance, TwistCovariance` arrays for having been initialized with `36` elements.
So here removing this check may require adding the initialization in [FROSOdometry](https://github.com/rapyuta-robotics/rclUE/blob/update_autogen/Source/rclUE/Public/Msgs/ROS2OdometryMsg.h)?

Source/RapyutaSimulationPlugins/Public/Core/RRConversionUtils.h
```diff
@@ -118,18 +113,24 @@ class URRConversionUtils : public UBlueprintFunctionLibrary
...
+    //                                       const double& InputY,
+    //                                       const double& InputZ,
+    //                                       double& OutputX,
+    //                                       double& OutputY,
+    //                                       double& OutputZ)
+    // {
+    //     OutputX = InputX * 100.f;
+    //     OutputY = -InputY * 100.f;
+    //     OutputZ = InputZ * 100.f;
+    // }
```
Tadinu:
```suggestion
#if RAPYUTA_SIM_DEBUG
    // This would be obviated in UE5, in which FVector is of double by default
    FORCEINLINE static void VectorROSToUE(const double& InputX,
                                          const double& InputY,
                                          const double& InputZ,
                                          double& OutputX,
                                          double& OutputY,
                                          double& OutputZ)
    {
        OutputX = InputX * 100.f;
        OutputY = -InputY * 100.f;
        OutputZ = InputZ * 100.f;
     }
#endif
```

Source/RapyutaSimulationPlugins/Public/Core/RRConversionUtils.h
```diff
@@ -168,16 +169,11 @@ class URRConversionUtils : public UBlueprintFunctionLibrary
...
         FROSOdometry Output = Input;
 
-        VectorROSToUE(Input.pose_pose_position_x,
-                      Input.pose_pose_position_y,
-                      Input.pose_pose_position_z,
-                      Output.pose_pose_position_x,
-                      Output.pose_pose_position_y,
-                      Output.pose_pose_position_z);
-        Output.pose_pose_orientation = QuatROSToUE(Output.pose_pose_orientation);
+        VectorROSToUE(Input.PosePosePosition, Output.PosePosePosition);
```
Tadinu, line 179: This should come with a note that `FROSOdometry` does not use `double` for position anymore for now but will be automatically resolved in UE5.

Tadinu (approved):
Thanks a lot for the heavy-lifting 🙏🏼 Such a sight to see all naming convention complied

Source/RapyutaSimulationPlugins/Public/Core/RRConversionUtils.h
yuokamoto: why resolved in UE5

Source/RapyutaSimulationPlugins/Private/Tools/RRROS2OdomPublisher.cpp
yuokamoto: Got it. that should be handled by UE_tools autogen.
I added limitation/todo of UE_tools readme

Source/RapyutaSimulationPlugins/Public/Core/RRConversionUtils.h
Tadinu: In UE5 FVector is [double](https://docs.unrealengine.com/5.0/en-US/large-world-coordinates-in-unreal-engine-5/#:~:text=Engine%205%2C%20the-,FVector,-casts%20will%20continue) by default

PR #106, opened 2022-11-10
ServerSpawnEntity: adjust location but always spawn

tetorea:
In practice, robots may not spawn frequently because of obstacles, just add the flag to always spawn

### DIFF
--- a/Source/RapyutaSimulationPlugins/Private/Tools/SimulationState.cpp
+++ b/Source/RapyutaSimulationPlugins/Private/Tools/SimulationState.cpp
@@ -320,7 +320,8 @@ AActor* ASimulationState::ServerSpawnEntity(const FROSSpawnEntityReq& InROSSpawn
     }
 
     // SpawnActorDeferred to set parameters beforehand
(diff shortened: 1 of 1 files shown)


Tadinu (approved):
Should be fine with a note that, the actual entity's transform could be different from one specified in `SpawnEntity`, thus needs to inform ros side to get synchronized with it. And that's why it is rejected to be spawned upon collision at the beginning.

tetorea:
> Should be fine with a note that, the actual entity's transform could be different from one specified in `SpawnEntity`, thus needs to inform ros side to get synchronized with it. And that's why it is rejected to be spawned upon collision at the beginning.

I see, then a better option would be to add the ability to choose with a member variable, for example?

Tadinu:
@yuokamoto just confirmed with me that it's ros/client side's responsibility to keep being informed about robot's pose in realtime, so it's fine I guess, just might we need to issue a note to users through the message in SpawnEntity Srv's Response, upon a comparison check with input pose.

arturocuma (approved):

PR #109, opened 2022-11-11
change postinitialization to preinitialization

yuokamoto:


Source/RapyutaSimulationPlugins/Private/Robots/RRBaseRobot.cpp
```diff
@@ -63,7 +64,7 @@ void ARRBaseRobot::PostInitializeComponents()
     }
 
     // which does the possessing, thus must be called afterwards
```
Tadinu, line 66: This comment is thus not warranted anymore, which is only for Post~

Tadinu (approved):

Source/RapyutaSimulationPlugins/Private/Robots/RRBaseRobot.cpp
```diff
@@ -63,7 +64,7 @@ void ARRBaseRobot::PostInitializeComponents()
     }
```
tetorea, line 64: Need to change the log text few lines above which still uses `PostInitializeComponents`

tetorea (approved):

PR #112, opened 2022-11-17
fix clang format error. remove blck_all collision setting to use with…

yuokamoto:
… InteractiveMarker

### DIFF
--- a/Source/RapyutaSimulationPlugins/Private/Core/RRProceduralMeshComponent.cpp
+++ b/Source/RapyutaSimulationPlugins/Private/Core/RRProceduralMeshComponent.cpp
@@ -68,27 +68,23 @@ bool URRProceduralMeshComponent::InitializeMesh(const FString& InMeshFileName)
             else
             {
                 URRThreadUtils::DoAsyncTaskInThread<void>(
-                    [this, InMeshFileName]()
-                    {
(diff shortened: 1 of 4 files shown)


Source/RapyutaSimulationPlugins/Private/Core/RRProceduralMeshComponent.cpp
```diff
@@ -68,27 +68,23 @@ bool URRProceduralMeshComponent::InitializeMesh(const FString& InMeshFileName)
             else
             {
                 URRThreadUtils::DoAsyncTaskInThread<void>(
-                    [this, InMeshFileName]()
-                    {
+                    [this, InMeshFileName]() {
```
Tadinu, line 71: This seems to make inline `{` only in lambda, but not for if/else, so abit inconsistent but more essentially, just curious why the change?

Tadinu (approved):

Source/RapyutaSimulationPlugins/Private/Core/RRProceduralMeshComponent.cpp
yuokamoto: hmm, not sure it was applied by clang-format of vscode.

PR #117, opened 2022-11-29
Crowd Float movement: Handling for Non-deceleration + 2D movement

Tadinu:
`URRFloatingMovementComponent`: Override
* `IsExceedingMaxSpeed(float InMaxSpeed)` to handle 2D movement vel
* `GetPenetrationAdjustment(const FHitResult& InHit)` to allow for custom PenetrationPullbackDistance config, which is not exposed from [UMovementComponent](https://github.com/EpicGames/UnrealEngine/blob/release/Engine/Source/Runtime/Engine/Private/Components/MovementComponent.cpp#L595)
* Add `bUseDecelerationForPaths, b2DMovement, PenetrationPullbackDistance`

`URRCrowdFollowingComponent`
* Add `FloatMovementComp, MaxCrowdSpeed`
* Override `FollowPathSegment(float InDeltaTime)` to keep velocity's magnitude to MaxCrowdSpeed if `FloatMovementComp` does not use `bUseDecelerationForPaths`
* Override `OnPathFinished()` to turn off `bStopMovementOnFinish`, which is force-turned ON in `UPathFollowingComponent::RequestMove` https://github.com/EpicGames/UnrealEngine/blob/release/Engine/Source/Runtime/AIModule/Private/Navigation/PathFollowingComponent.cpp#L363


Source/RapyutaSimulationPlugins/Public/Core/RRGeneralUtils.h
```diff
@@ -294,4 +294,56 @@ class RAPYUTASIMULATIONPLUGINS_API URRGeneralUtils : public UBlueprintFunctionLi
...
     }
+
+    /**
+     * @brief Check if a velocity's magnitude exceeds a given max speed
+     *
+     * @param InVel
+     * @param InMaxSpeed
+     * @param b2DMovement
+     */
+    FORCEINLINE static bool IsVelocityExceedingMaxSpeed(const FVector& InVel, float InMaxSpeed, bool b2DMovement)
```
yuokamoto: Can we genelize this as IsVecotorExceedMax?

Source/RapyutaSimulationPlugins/Public/Core/RRGeneralUtils.h
```diff
@@ -294,4 +294,56 @@ class RAPYUTASIMULATIONPLUGINS_API URRGeneralUtils : public UBlueprintFunctionLi
...
+    }
+
+    /**
+     * @brief Set a given velocity's magnitude to its max allowed speed
+     *
+     * @param InVel
+     * @param InMaxSpeed
+     * @param b2DMovement
+     */
+    FORCEINLINE static void SetVelocityToMaxSpeed(FVector& InVel, float InMaxSpeed, bool b2DMovement)
```
yuokamoto: Can we rename to Vector instead of Velocity?

yuokamoto (commented):
lgtm
I think functions for Velocity can be used for Vectors as well. it maybe better to rename those.

Source/RapyutaSimulationPlugins/Public/Core/RRGeneralUtils.h
Tadinu: I also thought about it as a util in `URRMathUtils` instead. If we do so, since it would be not very nice as writing a `foo(FVector, MaxMagnitude, b2DVector)` so it should be a template-based one like:
```
template<typename T>
static bool IsVectorExceeedingMaxMagnitude(const T& InVector, float InMaxMagnitude)
```
and call UE's built-in `SizeSquared()` for either T being `FVector` or `FVector2D`.
However, there are 2 issues with it:
* Using it would make us have to create an automatic var `FVector2D` from the input `Velocity`, before passing it to the util
* `b2DMovement` cannot be a parameter (due to its already inferred from typename T)

So yes, we could but then the util itself would not be very essential since `Velocity` is inherently always a 3D `FVector`. I haven't found a better idea yet tho.

Source/RapyutaSimulationPlugins/Public/Core/RRGeneralUtils.h
Tadinu: Same reason as above but then since Velocity is always a 3d-vector only, and basically we also want to provide `b2DMovement` as Velocity-specific param

Source/RapyutaSimulationPlugins/Public/Core/RRGeneralUtils.h
yuokamoto: I don't think you need to make it template but just for FVector.

Tadinu: Sure, will do but I still think adding `b2D` for a util function having `FVector` does not sound very generic, for the idea of generalizing. Someone may wonder why not using template, so it's kinda half-way generalizing maybe

yuokamoto (approved):

PR #119, opened 2022-12-06
update tb param, make set param accessible from bp.

yuokamoto:


Source/RapyutaSimulationPlugins/Public/Robots/RRRobotROS2Interface.h
```diff
@@ -116,8 +116,10 @@ class RAPYUTASIMULATIONPLUGINS_API URRRobotROS2Interface : public UObject
     /**
      * @brief Setup ROS Params, overridable by child classes to config custom ROS2 Interface's params
      */
-    UFUNCTION()
-    virtual void SetupROSParams();
+    UFUNCTION(BlueprintNativeEvent, BlueprintCallable)
+    void SetupROSParams();
```
Tadinu, line 120: This is a virtual function, meaning it is currently re-implemented by various other classes, so it should be kept.
For BP implementation support, it'd be better to create another one calling this function.

Source/RapyutaSimulationPlugins/Public/Robots/RRRobotROS2Interface.h
```diff
@@ -116,8 +116,10 @@ class RAPYUTASIMULATIONPLUGINS_API URRRobotROS2Interface : public UObject
     /**
      * @brief Setup ROS Params, overridable by child classes to config custom ROS2 Interface's params
      */
-    UFUNCTION()
-    virtual void SetupROSParams();
+    UFUNCTION(BlueprintNativeEvent, BlueprintCallable)
+    void SetupROSParams();
+
+    virtual void SetupROSParams_Implementation();
```
Tadinu:
```suggestion
    UFUNCTION()
    virtual void SetupROSParams();

    UFUNCTION(BlueprintNativeEvent, BlueprintCallable)
    void BPSetupROSParams();
```
So we could keep various child cpp classes's API header without having to update them with `_Implementation`, which is BP coupling.

Source/RapyutaSimulationPlugins/Private/Robots/RRRobotROS2Interface.cpp
```diff
@@ -16,13 +16,8 @@
...
 
-void URRRobotROS2Interface::SetupROSParams()
+void URRRobotROS2Interface::SetupROSParams_Implementation()
 {
-    bPublishOdom = true;
-    bPublishOdomTf = false;
-    CmdVelTopicName = TEXT("cmd_vel");
-    JointsCmdTopicName = TEXT("joint_states");
-    bWarnAboutMissingLink = true;
 }
```
Tadinu: Better to keep these to be reusable by child cpp classes?

yuokamoto: - those are set as default value in header already.
- if child class do not override this function, parameter become this values even if you changes member value in child class

Source/RapyutaSimulationPlugins/Public/Robots/RRRobotROS2Interface.h
yuokamoto: it is used only by SBROS2Interface.

Tadinu: Yeah it is used by Skeletal robot also, but I get the idea to make this supported by child BP.
Since this is one of the core class, ideally I'd still prefer an agnostic style of API, so either child cpp or BP class could use, but it might not be fully supported by UE.

Tadinu (approved):

PR #121, opened 2022-12-19
RRRobotModelInfo add WholeBodyStaticMeshName

Tadinu:
* RRRobotModelInfo add WholeBodyStaticMeshName, read by RRSDFParser due to only being used for .world file
* RRStaticMeshComponent::InitializeMesh() use static mesh from InMeshFilePath if exist

### DIFF
--- a/Source/RapyutaSimulationPlugins/Private/Core/RRSDFParser.cpp
+++ b/Source/RapyutaSimulationPlugins/Private/Core/RRSDFParser.cpp
@@ -261,7 +261,15 @@ bool FRRSDFParser::ParseModelUESpecifics(const sdf::ElementPtr& InModelElement, 
         endEffectorElement = endEffectorElement->GetNextElement(SDF_ELEMENT_END_EFFECTOR);
     }
 
-    // 6- WholeBody's material
+    // 6- WholeBody's static mesh (for World model only to make use of pre-baked UE static mesh)
+    sdf::ElementPtr staticMeshElement = ueElement->FindElement(SDF_ELEMENT_LINK_STATIC_MESH);
+    if (staticMeshElement)
+    {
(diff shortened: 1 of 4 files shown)


Source/RapyutaSimulationPlugins/Private/Core/RRSDFParser.cpp
```diff
@@ -261,7 +261,15 @@ bool FRRSDFParser::ParseModelUESpecifics(const sdf::ElementPtr& InModelElement,
         endEffectorElement = endEffectorElement->GetNextElement(SDF_ELEMENT_END_EFFECTOR);
     }
 
-    // 6- WholeBody's material
+    // 6- WholeBody's static mesh (for World model only to make use of pre-baked UE static mesh)
+    sdf::ElementPtr staticMeshElement = ueElement->FindElement(SDF_ELEMENT_LINK_STATIC_MESH);
```
yuokamoto, line 265: is assumption input xml has only one static mesh?
Please write those in somewhere when you write doc

Tadinu: So sdf itself actually already provides multi links, each of which comes with a mesh of it's own, but they are meant for ones loaded into runtime mesh in UE. (1)
* Thus, the idea of a whole-body single mesh is mainly useful only for large models that we want to utilize the pre-baked static asset and loaded into Sim as static entity as a whole, like in this case for world entity.
* We could provide multi-static meshes support, but generally, our intention is also to avoid relying on static assets too much due to their burden brought to the package.
* So ultimately, for multi-mesh models, we should load them directly from CAD instead as already provided in (1).
I will add those to doc

yuokamoto (approved):

PR #125, opened 2022-12-30
Warehouse + DataSynth utils

Tadinu:
The branch has been developed for a while mostly for the warehouse data synth app and needs to be merged to save maintenance burden & avoid ongoing conflicts with mainstream dev. It contains the following warehouse-related utils:
* Dynamic texture loading from HDR & object coating at runtime, which is used for pallets & boxes.
* Dynamic Level streaming at run time, which is gonna be used to load arbitrary photorealistic warehouse maps by @StefanIvanovTA to `RapyutaGeneric`
* Dynamic realistic light profile/IES loading
* `FRRActorSpawnInfo` add bIsOverlapEventEnabled;RRProceduralMeshComponent, RRStaticMeshComponent pass it to EnableOverlapping()

`ARRMeshActor`
* Segmentation mask value assigning using CutomDepthStencils, currently hardcoded due to incorrect pixel capturing in UE5
* Bulk-homogeneous-entities segmask value assigning (like a stack of pallets as in a common segmentation)

### DIFF
--- a/Source/RapyutaSimulationPlugins/Private/Core/RRActorCommon.cpp
+++ b/Source/RapyutaSimulationPlugins/Private/Core/RRActorCommon.cpp
@@ -210,36 +210,20 @@ void URRActorCommon::OnStartSim()
 
 void URRActorCommon::SetupEnvironment()
 {
-    // Here, we initialize artifacts that utilized play resources
     UWorld* currentWorld = GetWorld();
     checkf(currentWorld, TEXT("[URRActorCommon::SetupEnvironment] Failed fetching Game World"));
 
-    // Fetch Main background environment
-    if (!MainEnvironment)
-    {
-        MainEnvironment = URRUObjectUtils::FindEnvironmentActor(currentWorld);
-        // Not all maps has MainEnvironment setup
-    }
+    // By default, these just reference GameState's floor & wall but could be overridable in child class setup
+    SceneFloor = GameState->MainFloor;
+    SceneWall = GameState->MainWall;
 
-    if (!MainFloor)
-    {
-        MainFloor = URRUObjectUtils::FindFloorActor(currentWorld);
-        // Not all maps has MainFloor setup
-    }
-
-    if (!MainWall)
-    {
-        MainWall = URRUObjectUtils::FindWallActor(currentWorld);
-        // Not all maps has MainWall setup
-    }
-
-    // Spawn MainCamera
-    MainCamera =
+    // Spawn Scene main camera
+    SceneCamera =
         Cast<ARRCamera>(URRUObjectUtils::SpawnSimActor(GetWorld(),
                                                        SceneInstanceId,
                                                        ARRCamera::StaticClass(),
                                                        TEXT("RapyutaCamera"),
-                                                       FString::Printf(TEXT("%d_MainCamera"), SceneInstanceId),
+                                                       FString::Printf(TEXT("%d_SceneCamera"), SceneInstanceId),
                                                        FTransform(SceneInstanceLocation),
                                                        ESpawnActorCollisionHandlingMethod::AdjustIfPossibleButAlwaysSpawn));
 }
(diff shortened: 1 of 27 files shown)


Source/RapyutaSimulationPlugins/Private/Core/RRActorCommon.cpp
```diff
@@ -205,36 +230,20 @@ void URRActorCommon::OnStartSim()
...
-        // Not all maps has MainFloor setup
-    }
-
-    if (!MainWall)
-    {
-        MainWall = URRUObjectUtils::FindWallActor(currentWorld);
-        // Not all maps has MainWall setup
-    }
+    // By default, these just reference GameState's floor & wall but could be overridable in child class setup
+    SceneFloor = GameState->MainFloor;
```
yuokamoto, line 237: where are these SecneFloor and SceneWall used?

Source/RapyutaSimulationPlugins/Private/Core/RRGameState.cpp
```diff
@@ -64,25 +67,53 @@ void ARRGameState::StartSim()
...
         UE_LOG(LogRapyutaCore, Display, TEXT("[ARRGameState]:: SIM SCENE INSTANCE [%d] STARTED SUCCESSFULLY! "), i);
     }
 }
 
+void ARRGameState::SetupEnvironment()
+{
+    FetchEnvStaticActors();
+}
+
+void ARRGameState::FetchEnvStaticActors()
```
yuokamoto, line 97: What is the purpose of fetching wall, floor actors?

Source/RapyutaSimulationPlugins/Private/Core/RRPlayerController.cpp
```diff
@@ -53,8 +53,12 @@ bool ARRPlayerController::Initialize()
     check(ActorCommon);
     check(ActorCommon->SceneInstanceId == SceneInstanceId);
 
-    MainCamera = ActorCommon->MainCamera;
-    check(MainCamera);
+    SceneCamera = ActorCommon->SceneCamera;
+    check(SceneCamera);
```
yuokamoto, line 57: Do we really need these check/verify, e.g. is it possible to change to ensure? Crashing editor take time to restarting

Source/RapyutaSimulationPlugins/Private/Core/RRUObjectUtils.cpp
```diff
@@ -12,6 +12,101 @@
 #include "Core/RRMathUtils.h"
 #include "Core/RRMeshActor.h"
 
+static TArray<uint8> DATA_SYNTH_CUSTOM_DEPTH_STENCILS = []()
+{
+    TArray<uint8> data = {
```
yuokamoto, line 19: what's this for?

yuokamoto (approved):
Honestly says, I can understand not all. I think most of the changes for data generation for machine learning and dynamic mesh spawning.

It is ok to merge if it won't break but I want to you update/create documentation for those features.

Source/RapyutaSimulationPlugins/Private/Core/RRActorCommon.cpp
Tadinu: `SceneFloor, SceneWall` are used [here](https://github.com/rapyuta-robotics/RapyutaDataSynth/blob/devel/Source/RapyutaSimDataSynth/Private/SceneCapture/RRCaptureDirector.cpp#L211) in case users want to randomize the background

Source/RapyutaSimulationPlugins/Private/Core/RRGameState.cpp
Tadinu: We need walls, and floor to be available (fetched from the map) in order for the first spawned pallet to detect its collision-free locations, which are used during pallet placement randomization

Source/RapyutaSimulationPlugins/Private/Core/RRPlayerController.cpp
Tadinu: We could though I'd also want to clarify the meanings of `ensure` vs `check/verify`
* `ensure` acts as exceptions that should come from incorrect usage, mostly to inform users about how to interact with Sim
* `check/verify` acts like `assert`, purely dev-oriented, thus are more essential to dev to prevent early critical bugs, which in this case are required since SceneCamera is just necessary for the scene capturing.
It might be not straightforward to decide which to use, generally `check/verify` helps save time in debugging, since there is no point to go further with a nullptr variant. But yeah for sake of saving and restarting, we should switch to `ensure`.

Tadinu: We could though I'd also want to clarify the meanings of `ensure` vs `check/verify`
* `ensure` acts as exceptions that should come from incorrect usage, mostly to inform users about how to interact with Sim
* `check/verify` acts like `assert`, purely dev-oriented, thus are more essential to dev to prevent early critical bugs, which in this case are required since SceneCamera is just necessary for the scene capturing.
It might be not straightforward to decide which to use, generally `check/verify` helps save time in debugging, since there is no point to go further with a nullptr variant. But yeah for sake of saving athe restart, we should switch to `ensure`.

Source/RapyutaSimulationPlugins/Private/Core/RRUObjectUtils.cpp
Tadinu: They are hardcoded Segmentation mask values due to incorrect pixel capturing in UE5 (like the ones read from output png are different from the values written into the stencil buffer) as mentioned in the description

Tadinu: They are hardcoded Segmentation mask values due to incorrect pixel capturing in UE5 (like the ones read from output png are different from the values written into the custom depth buffer) as mentioned in the description

PR #126, opened 2023-01-19
added default entry map

larryng-rapyuta:
If this Entry map is set as the default game map, can select map to enter using -targetmap={MAP}, game mode for this target map can be selected by using -gamemode={GAME MODE}


Tadinu (approved):
Regarding game mode, it seems we could also use `GAME=`?
https://github.com/EpicGames/UnrealEngine/blob/release/Engine/Source/Runtime/Engine/Private/GameInstance.cpp#L1391

larryng-rapyuta:
we can but that wont set the gamemode for the map we are entering

yuokamoto (approved):

PR #134, opened 2023-02-28
AutoRegister user-specified BP Classes as Spawnable Entity types

Tadinu:
* ARRGameState add BPSpawnableClassNames + RegisterSpawnableBPClasses()
* URRAssetUtils Add FindBlueprintClass()
* ARRBaseRobot Requires static BP Robot to have its model name empty or prefixed with `BP`

### DIFF
--- a/Source/RapyutaSimulationPlugins/Private/Core/RRGameState.cpp
+++ b/Source/RapyutaSimulationPlugins/Private/Core/RRGameState.cpp
@@ -67,6 +67,9 @@ void ARRGameState::StartSim()
     UE_LOG_WITH_INFO(LogRapyutaCore, Display, TEXT("MAX SPLIT SCREEN PLAYERS: %d"), maxSplitscreenPlayers);
     verify(SCENE_INSTANCES_NUM <= maxSplitscreenPlayers);
 
+    // Register BP spawnable classes
+    RegisterSpawnableBPClasses();
+
     // 0- Stream level & Fetch static-env actors
     SetupEnvironment();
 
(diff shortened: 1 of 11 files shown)


Source/RapyutaSimulationPlugins/Private/Core/RRGameState.cpp
```diff
@@ -89,6 +92,19 @@ void ARRGameState::StartSim()
     }
 }
 
+void ARRGameState::RegisterSpawnableBPClasses()
```
yuokamoto: Shouldn't these logic be part of GameMode since GameMode has  SimulationState.

Source/RapyutaSimulationPlugins/Public/Core/RRAssetUtils.h
```diff
@@ -242,4 +242,45 @@ class RAPYUTASIMULATIONPLUGINS_API URRAssetUtils : public UBlueprintFunctionLibr
...
+    FORCEINLINE static UClass* FindBlueprintClass(const FString& InBlueprintClassName)
+    {
+        IAssetRegistry& assetRegistry = GetAssetRegistry();
+        if (assetRegistry.IsLoadingAssets())
+        {
+            assetRegistry.SearchAllAssets(true);
+        }
+
+        FString targetName = InBlueprintClassName;
+        targetName.RemoveFromEnd(TEXT("_C"), ESearchCase::CaseSensitive);
```
yuokamoto, line 256: whats _C means?

Source/RapyutaSimulationPlugins/Private/Core/RRGameState.cpp
Tadinu: yes, it should be server-only, so moved to SimulationState & invoked by ROS2GameMode

Source/RapyutaSimulationPlugins/Public/Core/RRAssetUtils.h
Tadinu: Yeah an odd affix generated by UE for Blueprint Class asset path to be loaded to UObject. typically seems no docs about it tho.
https://github.com/EpicGames/UnrealEngine/blob/release/Engine/Plugins/Animation/GameplayInsights/Source/GameplayInsights/Private/GameplayProvider.cpp#L28

yuokamoto (approved):

PR #135, opened 2023-03-06
fix init order to start ROS2 node later

yuokamoto:
Make sure to start ROS2 Node and services after all other settings are completed

### DIFF
--- a/Source/RapyutaSimulationPlugins/Private/Core/RRROS2GameMode.cpp
+++ b/Source/RapyutaSimulationPlugins/Private/Core/RRROS2GameMode.cpp
@@ -38,18 +38,15 @@ void ARRROS2GameMode::InitGame(const FString& InMapName, const FString& InOption
     UE_LOG_WITH_INFO(
         LogRapyutaCore, Display, TEXT("ShouldUseThreadingForPerformance: %d"), FApp::ShouldUseThreadingForPerformance());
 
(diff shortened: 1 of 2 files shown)


Source/RapyutaSimulationPlugins/Private/Core/RRGameMode.cpp
```diff
@@ -138,6 +138,9 @@ void ARRGameMode::StartPlay()
     // 2- START SIM
     StartSim();
 
+    // 3- START ROS2 Nodes
+    InitSim();
+
```
Tadinu: Is that `InitSim()` now already called in `ARRROS2GameMode::StartPlay()`, which happens after all resources loading?

yuokamoto: StartPlay is overwridden without calling Super::StartPlay() in RRGameMode

Tadinu: Yeah, it is still called [here](https://github.com/rapyuta-robotics/RapyutaSimulationPlugins/blob/devel/Source/RapyutaSimulationPlugins/Private/Core/RRGameMode.cpp#L203) as in a deferred style after all resources have been loaded.

Tadinu (approved):

PR #137, opened 2023-03-10
Asset utils: Fetch static UObject/UClass from asset in ctor

Tadinu:


yuokamoto (approved):
Just from curiosity, when we needs to resave assets?

yuokamoto (approved):
Just from curiosity, when we needs to resave assets?

Tadinu:
As observed, the warning of specific assets over their missing UE version (thus might not be fully functional) appears as being loaded from a map or directly from cpp, or opened in the Editor.
Normally UE should provide backward compatibility to assets but as transition from UE4 to UE5, it seems the project's Contents are not auto converted.

PR #140, opened 2023-03-22
RRAssetUtils add CreateUClass()

Tadinu:
* `UARRAssetUtils` add `CreateUClass()` creating blueprint & generate its class at runtime with an optional CDO modifier on the blueprint. An application example is to create Wheel BP from raw configurations at runtime.
* `FRRRobotWheelProperty` tune runtime wheel params

### DIFF
--- a/Source/RapyutaSimulationPlugins/Public/Core/RRAssetUtils.h
+++ b/Source/RapyutaSimulationPlugins/Public/Core/RRAssetUtils.h
@@ -8,17 +8,18 @@
 
 // UE
 #include "AssetRegistry/AssetRegistryModule.h"
-#include "AssetToolsModule.h"
 #include "Engine/ObjectLibrary.h"
 #include "Engine/StaticMesh.h"
 #include "Kismet/BlueprintFunctionLibrary.h"
-#if WITH_EDITOR
-#include "BlueprintCompilationManager.h"
-#include "Kismet2/KismetEditorUtilities.h"
-#endif
-#include "KismetCompilerModule.h"
 #include "Materials/Material.h"
 #include "PhysicalMaterials/PhysicalMaterial.h"
+
+#if WITH_EDITOR
+#include "AssetToolsModule.h"
+#include "BlueprintCompilationManager.h"
+#include "Kismet2/KismetEditorUtilities.h"
+#include "KismetCompilerModule.h"
+#endif
 
 // RapyutaSim
 #include "RapyutaSimulationPlugins.h"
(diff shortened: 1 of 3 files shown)


Source/RapyutaSimulationPlugins/Public/Robots/RRRobotStructs.h
```diff
@@ -901,17 +901,18 @@ struct RAPYUTASIMULATIONPLUGINS_API FRRRobotWheelProperty
         FRRRobotWheelProperty wheelProp(MoveTemp(InWheelName));
         // [Radius/Width]
         wheelProp.WheelRadius = bFrontWheel ? 13.f : (wheelProp.WheelName.Contains(TEXT("caster")) ? 9.f : 16.f);
-        wheelProp.WheelWidth = bFrontWheel ? 10.f : 20.f;
+        wheelProp.WheelWidth = bFrontWheel ? 10.f : (wheelProp.WheelName.Contains(TEXT("caster")) ? 25.f : 14.f);
```
yuokamoto: As we discussed, please move these structure/class/metho to RobotImporter since it is used only by them  or add comment that these are used in the future.
Since people only see this repo, it is a little bit confusing.

Tadinu: So I don't remember this being settled in discussion, the very reason why we put `FRRRobotWheelProperty` here is it is the host struct to store metadata defined in URDF/SDF, which is already part of RSPlugins as defined [here](https://github.com/rapyuta-robotics/RapyutaSimulationPlugins/blob/devel/Source/RapyutaSimulationPlugins/Public/Core/RRURDFParser.h#L50).
* As such, we could move urdf/sdf **back** to Robotimporter, but given the very fact that URDF/SDF is not only to describe robot but any mesh-based entity, not sure we should tho.
* For now `FRRRobotWheelProperty` though it's only used by RobotImporter, but all of its properties are taken from `UChaosVehicleWheel` actually, so generally users could use it for creating their own custom vehicle wheel at runtime.

Source/RapyutaSimulationPlugins/Public/Core/RRAssetUtils.h
```diff
@@ -309,4 +321,94 @@ class RAPYUTASIMULATIONPLUGINS_API URRAssetUtils : public UBlueprintFunctionLibr
...
         return foundClass;
     }
+
+    /**
+     * @brief Create a child UClass from parent UClass
+     * @param InParentClass
+     * @param InCDOFunc
+     * @return UClass*
+     */
+    // Ref: [FKismetEditorUtilities::CreateBlueprintFromClass()], [FBlueprintEditorUtils::PostDuplicateBlueprint()]
```
yuokamoto: Please add link to the doc

chalice-graeme (approved):

yuokamoto (approved):

PR #149, opened 2023-04-13
Add spawnable

yuokamoto:
- Add Utilities to RRJointComponent
- Add BP interface to RRGeneralUtils
- Add Naive implementation to find shortest angle difference


Source/RapyutaSimulationPlugins/Private/Drives/RRJointComponent.cpp
```diff
@@ -48,6 +48,11 @@ void URRJointComponent::SetPoseTarget(const FVector& InPosition, const FRotator&
     OrientationTarget = InOrientation;
 };
 
+bool URRJointComponent::IsReachPoseTarget(const float InTolerance)
```
Tadinu:
```suggestion
bool URRJointComponent::HasReachedPoseTarget(const float InTolerance)
```

Source/RapyutaSimulationPlugins/Public/Core/RRMathUtils.h
```diff
@@ -24,13 +24,38 @@ class RAPYUTASIMULATIONPLUGINS_API URRMathUtils : public UBlueprintFunctionLibra
...
+            {
+                angleDiff -= 360;
+            }
+            else if (angleDiff < -180)
+            {
+                angleDiff += 360;
+            }
+        }
+        return angleDiff;
+    }
```
Tadinu: Could use [FRotator::NormalizeAxis(angleDiff)](https://github.com/EpicGames/UnrealEngine/blob/release/Engine/Source/Runtime/Core/Public/Math/Rotator.h#L416)?

Tadinu (approved):
Just some minor comments, btw I suppose all of those clang-format incompatibilities have been resolved, unclear why still once in a while it appears?

Source/RapyutaSimulationPlugins/Public/Core/RRMathUtils.h
yuokamoto: Thanks NormalizeAxis works. I removed.

yuokamoto:
> Just some minor comments, btw I suppose all of those clang-format incompatibilities have been resolved, unclear why still once in a while it appears?

Did you install pre-commit hook? then it is formated when you commit

PR #154, opened 2023-05-01
Refactor check to decouple with RR~ game framework

yuokamoto:
- update dependency and check. 
- Make ROS2GameMode and ARRBaseRobot available without RRGameSingleton and etc
- Rem use of `ARRGameInstance` which is just printing log
- Decoupling with `ARRPlayerController`, which is only used with SceneDirector
- update log level
- Make RAPYUTA_USE_SCENE_DIRECTOR optional
- Add RAPYUTA_SIM_VERBOSE

### DIFF
--- a/Source/RapyutaSimulationPlugins/Private/Core/RRBaseActor.cpp
+++ b/Source/RapyutaSimulationPlugins/Private/Core/RRBaseActor.cpp
@@ -45,17 +45,11 @@ void ARRBaseActor::PreInitializeComponents()
     // SETUP + CONFIGURE GAME & SIM COMMON OBJECTS
     // NOTE: These are used ONLY for dynamic runtime actors.
     // In child BP actors, except for GameSingleton, other Game framework entities maybe not available yet
-#if RAPYUTA_USE_SCENE_DIRECTOR
+
+    // pointers for convinence
     RRGameState = URRCoreUtils::GetGameState<ARRGameState>(this);
-    check(RRGameState)
-#endif
-        RRROS2GameMode = URRCoreUtils::GetGameMode<ARRROS2GameMode>(this);
-
-    if (IsNetMode(NM_Standalone))
-    {
-        RRPlayerController = URRCoreUtils::GetPlayerController<ARRPlayerController>(SceneInstanceId, this);
-    }
-
+    RRPlayerController = URRCoreUtils::GetPlayerController<ARRPlayerController>(SceneInstanceId, this);
+    RRGameMode = URRCoreUtils::GetGameMode<ARRGameMode>(this);
     RRGameSingleton = URRGameSingleton::Get();
     if (RRGameSingleton == nullptr)
     {
--- a/Source/RapyutaSimulationPlugins/Private/Core/RRBaseActor.cpp
+++ b/Source/RapyutaSimulationPlugins/Private/Core/RRBaseActor.cpp
@@ -42,17 +42,25 @@ void ARRBaseActor::PreInitializeComponents()
 {
     Super::PreInitializeComponents();
 
-    // SETUP + CONFIGURE ESSENTIAL GAME & SIM COMMON OBJECTS
+    // SETUP + CONFIGURE GAME & SIM COMMON OBJECTS
     // NOTE: These are used ONLY for dynamic runtime actors.
     // In child BP actors, except for GameSingleton, other Game framework entities maybe not available yet
 #if RAPYUTA_USE_SCENE_DIRECTOR
-    GameState = URRCoreUtils::GetGameState<ARRGameState>(this);
-    check(GameState)
+    RRGameState = URRCoreUtils::GetGameState<ARRGameState>(this);
+    check(RRGameState)
 #endif
-    GameSingleton = URRGameSingleton::Get();
-    if(GameSingleton==nullptr)
+        RRROS2GameMode = URRCoreUtils::GetGameMode<ARRROS2GameMode>(this);
+
+    if (IsNetMode(NM_Standalone))
     {
-         UE_LOG_WITH_INFO_NAMED(LogRapyutaCore, Display, TEXT("GameSingleton is not child class of URRGameSingleton. Cannot use Asset loading."));
+        RRPlayerController = URRCoreUtils::GetPlayerController<ARRPlayerController>(SceneInstanceId, this);
+    }
+
+    RRGameSingleton = URRGameSingleton::Get();
+    if (RRGameSingleton == nullptr)
+    {
+        UE_LOG_WITH_INFO_NAMED(
+            LogRapyutaCore, Display, TEXT("GameSingleton is not child class of URRGameSingleton. Cannot use Asset loading."));
     }
 }
 
--- a/Source/RapyutaSimulationPlugins/Private/Core/RRBaseActor.cpp
+++ b/Source/RapyutaSimulationPlugins/Private/Core/RRBaseActor.cpp
@@ -5,7 +5,6 @@
 #include "Core/RRActorCommon.h"
 #include "Core/RRCoreUtils.h"
 #include "Core/RRGameMode.h"
-#include "Core/RRGameSingleton.h"
 #include "Core/RRGameState.h"
 #include "Core/RRPlayerController.h"
 #include "Core/RRUObjectUtils.h"
@@ -46,16 +45,15 @@ void ARRBaseActor::PreInitializeComponents()
     // SETUP + CONFIGURE ESSENTIAL GAME & SIM COMMON OBJECTS
     // NOTE: These are used ONLY for dynamic runtime actors.
     // In child BP actors, except for GameSingleton, other Game framework entities maybe not available yet
-    GameMode = URRCoreUtils::GetGameMode<ARRGameMode>(this);
+#if RAPYUTA_USE_SCENE_DIRECTOR
     GameState = URRCoreUtils::GetGameState<ARRGameState>(this);
-
-    if (IsNetMode(NM_Standalone))
-    {
-        PlayerController = URRCoreUtils::GetPlayerController<ARRPlayerController>(SceneInstanceId, this);
-    }
-
+    check(GameState)
+#endif
     GameSingleton = URRGameSingleton::Get();
-    check(GameSingleton);
+    if(GameSingleton==nullptr)
+    {
+         UE_LOG_WITH_INFO_NAMED(LogRapyutaCore, Display, TEXT("GameSingleton is not child class of URRGameSingleton. Cannot use Asset loading."));
+    }
 }
 
 bool ARRBaseActor::Initialize()
--- a/Source/RapyutaSimulationPlugins/Private/Core/RRGameInstance.cpp
+++ b/Source/RapyutaSimulationPlugins/Private/Core/RRGameInstance.cpp
@@ -11,28 +11,36 @@
 FString URRGameInstance::SMapName;
 void URRGameInstance::PreloadContentForURL(FURL InURL)
 {
+#if RAPYUTA_SIM_DEBUG
     UE_LOG(LogRapyutaCore, Display, TEXT("Preload Content for Map [%s]"), *InURL.Map)
+#endif
     Super::PreloadContentForURL(InURL);
 }
 
 AGameModeBase* URRGameInstance::CreateGameModeForURL(FURL InURL, UWorld* InWorld)
 {
-    UE_LOG_WITH_INFO(LogRapyutaCore, Display, TEXT(" URL MAP PATH [%s] loaded into WORLD [%s]"), *InURL.Map, *InWorld->GetName());
 
     SMapName = InWorld->GetName();
+#if RAPYUTA_SIM_DEBUG
+    UE_LOG_WITH_INFO(LogRapyutaCore, Display, TEXT(" URL MAP PATH [%s] loaded into WORLD [%s]"), *InURL.Map, *InWorld->GetName());
     URRCoreUtils::ScreenMsg(FColor::Yellow, FString::Printf(TEXT("MAP: %s"), *SMapName));
+#endif
     return Super::CreateGameModeForURL(InURL, InWorld);
 }
 
 void URRGameInstance::StartGameInstance()
 {
+#if RAPYUTA_SIM_DEBUG
     UE_LOG_WITH_INFO(LogRapyutaCore, Display, TEXT("StartGameInstance!"));
+#endif
     Super::StartGameInstance();
 }
 
 void URRGameInstance::Init()
 {
+#if RAPYUTA_SIM_DEBUG
     UE_LOG_WITH_INFO(LogRapyutaCore, Display, TEXT("INIT!"))
+#endif
     Super::Init();
     URRMathUtils::InitializeRandomStream();
 }
@@ -40,11 +48,15 @@ void URRGameInstance::Init()
 void URRGameInstance::LoadComplete(const float InLoadTime, const FString& InMapName)
 {
     Super::LoadComplete(InLoadTime, InMapName);
+#if RAPYUTA_SIM_DEBUG
     UE_LOG_WITH_INFO(LogRapyutaCore, Display, TEXT("[%s] LOADING COMPLETED!"), *InMapName);
+#endif
 }
 
 void URRGameInstance::OnStart()
 {
+#if RAPYUTA_SIM_DEBUG
     UE_LOG_WITH_INFO(LogRapyutaCore, Display, TEXT("ON START!"))
+#endif
     Super::OnStart();
 }
--- a/Source/RapyutaSimulationPlugins/Private/Core/RRGameInstance.cpp
+++ b/Source/RapyutaSimulationPlugins/Private/Core/RRGameInstance.cpp
@@ -11,7 +11,7 @@
 FString URRGameInstance::SMapName;
 void URRGameInstance::PreloadContentForURL(FURL InURL)
 {
-#if RAPYUTA_SIM_DEBUG
+#if RAPYUTA_SIM_VERBOSE
     UE_LOG(LogRapyutaCore, Display, TEXT("Preload Content for Map [%s]"), *InURL.Map)
 #endif
     Super::PreloadContentForURL(InURL);
@@ -19,9 +19,8 @@ void URRGameInstance::PreloadContentForURL(FURL InURL)
 
 AGameModeBase* URRGameInstance::CreateGameModeForURL(FURL InURL, UWorld* InWorld)
 {
-
     SMapName = InWorld->GetName();
-#if RAPYUTA_SIM_DEBUG
+#if RAPYUTA_SIM_VERBOSE
     UE_LOG_WITH_INFO(LogRapyutaCore, Display, TEXT(" URL MAP PATH [%s] loaded into WORLD [%s]"), *InURL.Map, *InWorld->GetName());
     URRCoreUtils::ScreenMsg(FColor::Yellow, FString::Printf(TEXT("MAP: %s"), *SMapName));
 #endif
@@ -30,7 +29,7 @@ AGameModeBase* URRGameInstance::CreateGameModeForURL(FURL InURL, UWorld* InWorld
 
 void URRGameInstance::StartGameInstance()
 {
-#if RAPYUTA_SIM_DEBUG
+#if RAPYUTA_SIM_VERBOSE
     UE_LOG_WITH_INFO(LogRapyutaCore, Display, TEXT("StartGameInstance!"));
 #endif
     Super::StartGameInstance();
@@ -38,7 +37,7 @@ void URRGameInstance::StartGameInstance()
 
 void URRGameInstance::Init()
 {
-#if RAPYUTA_SIM_DEBUG
+#if RAPYUTA_SIM_VERBOSE
     UE_LOG_WITH_INFO(LogRapyutaCore, Display, TEXT("INIT!"))
 #endif
     Super::Init();
@@ -48,14 +47,14 @@ void URRGameInstance::Init()
 void URRGameInstance::LoadComplete(const float InLoadTime, const FString& InMapName)
 {
     Super::LoadComplete(InLoadTime, InMapName);
-#if RAPYUTA_SIM_DEBUG
+#if RAPYUTA_SIM_VERBOSE
     UE_LOG_WITH_INFO(LogRapyutaCore, Display, TEXT("[%s] LOADING COMPLETED!"), *InMapName);
 #endif
 }
 
 void URRGameInstance::OnStart()
 {
-#if RAPYUTA_SIM_DEBUG
+#if RAPYUTA_SIM_VERBOSE
     UE_LOG_WITH_INFO(LogRapyutaCore, Display, TEXT("ON START!"))
 #endif
     Super::OnStart();
--- a/Source/RapyutaSimulationPlugins/Private/Core/RRGameMode.cpp
+++ b/Source/RapyutaSimulationPlugins/Private/Core/RRGameMode.cpp
@@ -17,14 +17,16 @@
 
 ARRGameMode::ARRGameMode()
 {
+#if RAPYUTA_USE_SCENE_DIRECTOR
     GameStateClass = ARRGameState::StaticClass();
     PlayerControllerClass = ARRPlayerController::StaticClass();
+#endif
 }
 
 void ARRGameMode::PreInitializeComponents()
 {
 #if RAPYUTA_SIM_DEBUG
-    UE_LOG_WITH_INFO(LogRapyutaCore, Warning, TEXT("PreInitializeComponents()"))
+    UE_LOG_WITH_INFO(LogRapyutaCore, Verbose, TEXT("PreInitializeComponents()"))
 #endif
     Super::PreInitializeComponents();
 }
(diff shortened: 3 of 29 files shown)


Source/RapyutaSimulationPlugins/Private/Core/RRBaseActor.cpp
```diff
@@ -46,16 +45,15 @@ void ARRBaseActor::PreInitializeComponents()
...
-
-    if (IsNetMode(NM_Standalone))
+    check(GameState)
+#endif
+    GameSingleton = URRGameSingleton::Get();
+    if(GameSingleton==nullptr)
     {
-        PlayerController = URRCoreUtils::GetPlayerController<ARRPlayerController>(SceneInstanceId, this);
+         UE_LOG_WITH_INFO_NAMED(LogRapyutaCore, Display, TEXT("GameSingleton is not child class of URRGameSingleton. Cannot use Asset loading."));
     }
```
Tadinu: `GameMode, GameState, GameSingleton, PlayerController` are optional  meaning they are not necessarily of our `ARR~` types but should all be fetched regardless of `RAPYUTA_USE_SCENE_DIRECTOR`.
So here you could just remove `check(GameState), check(GameSingleton), etc.`

Source/RapyutaSimulationPlugins/Private/Core/RRGameInstance.cpp
```diff
@@ -11,40 +11,52 @@
 FString URRGameInstance::SMapName;
 void URRGameInstance::PreloadContentForURL(FURL InURL)
 {
+#if RAPYUTA_SIM_DEBUG
     UE_LOG(LogRapyutaCore, Display, TEXT("Preload Content for Map [%s]"), *InURL.Map)
+#endif
```
Tadinu: This log is still informative for its information regarding what happens during init process + map name

Source/RapyutaSimulationPlugins/Private/Core/RRGameInstance.cpp
```diff
@@ -11,40 +11,52 @@
...
 
 AGameModeBase* URRGameInstance::CreateGameModeForURL(FURL InURL, UWorld* InWorld)
 {
-    UE_LOG_WITH_INFO(LogRapyutaCore, Display, TEXT(" URL MAP PATH [%s] loaded into WORLD [%s]"), *InURL.Map, *InWorld->GetName());
 
     SMapName = InWorld->GetName();
+#if RAPYUTA_SIM_DEBUG
+    UE_LOG_WITH_INFO(LogRapyutaCore, Display, TEXT(" URL MAP PATH [%s] loaded into WORLD [%s]"), *InURL.Map, *InWorld->GetName());
     URRCoreUtils::ScreenMsg(FColor::Yellow, FString::Printf(TEXT("MAP: %s"), *SMapName));
+#endif
```
Tadinu: Same reason above, Useful log

Source/RapyutaSimulationPlugins/Private/Core/RRGameInstance.cpp
```diff
@@ -11,40 +11,52 @@
...
     return Super::CreateGameModeForURL(InURL, InWorld);
 }
 
 void URRGameInstance::StartGameInstance()
 {
+#if RAPYUTA_SIM_DEBUG
     UE_LOG_WITH_INFO(LogRapyutaCore, Display, TEXT("StartGameInstance!"));
+#endif
```
Tadinu: Useful log

Source/RapyutaSimulationPlugins/Private/Core/RRGameInstance.cpp
```diff
@@ -11,40 +11,52 @@
...
     Super::Init();
     URRMathUtils::InitializeRandomStream();
 }
 
 void URRGameInstance::LoadComplete(const float InLoadTime, const FString& InMapName)
 {
     Super::LoadComplete(InLoadTime, InMapName);
+#if RAPYUTA_SIM_DEBUG
     UE_LOG_WITH_INFO(LogRapyutaCore, Display, TEXT("[%s] LOADING COMPLETED!"), *InMapName);
+#endif
```
Tadinu: Useful log signifying GameInstance having been fully loaded

Source/RapyutaSimulationPlugins/Private/Core/RRGameMode.cpp
```diff
@@ -17,61 +17,63 @@
...
 #endif
     Super::InitGameState();
 }
 
 void ARRGameMode::PrintSimConfig() const
 {
-    UE_LOG_WITH_INFO(LogRapyutaCore, Display, TEXT("SIM GAME MODE CONFIG -----------------------------"));
-    UE_LOG_WITH_INFO(LogRapyutaCore, Display, TEXT("- bBenchmark: %d"), bBenchmark);
+    UE_LOG_WITH_INFO(LogRapyutaCore, Verbose, TEXT("SIM GAME MODE CONFIG -----------------------------"));
+    UE_LOG_WITH_INFO(LogRapyutaCore, Verbose, TEXT("- bBenchmark: %d"), bBenchmark);
```
Tadinu: Useful log

Source/RapyutaSimulationPlugins/Private/Core/RRGameMode.cpp
```diff
@@ -115,8 +117,11 @@ void ARRGameMode::StartPlay()
     PrintSimConfig();
     PrintUEPreprocessors();
 
-    GameInstance = Cast<URRGameInstance>(GetGameInstance());
-    verify(GameInstance);
+    if(URRGameSingleton::Get()==nullptr)
```
Tadinu: clang-format

Source/RapyutaSimulationPlugins/Private/Core/RRGameMode.cpp
```diff
@@ -149,7 +154,11 @@ void ARRGameMode::StartSim()
     URRCoreUtils::LoadImageWrapperModule();
 
     // 2- LOAD SIM STATIC GLOBAL RESOURCES --
-    URRGameSingleton::Get()->InitializeResources();
+    auto* gameSingleton = URRGameSingleton::Get();
+    if(gameSingleton)
```
Tadinu: clang-format?

Source/RapyutaSimulationPlugins/Private/Core/RRGameMode.cpp
```diff
@@ -167,27 +176,33 @@ bool ARRGameMode::TryStartingSim()
...
-            UE_LOG_WITH_INFO(LogRapyutaCore, Fatal, TEXT("DYNAMIC RESOURCES LOADING TIMEOUT -> SHUTTING DOWN THE SIM..."))
-        },
-        sBeginTime,
-        ARRGameMode::SIM_START_TIMEOUT_SECS);
-
-    // Either resources have yet to be fully loaded or time out!
-    if (!bResult)
+    
+    auto* gameSingleton = URRGameSingleton::Get();
+    if(gameSingleton)
```
Tadinu: clang-format

Source/RapyutaSimulationPlugins/Private/Core/RRGameState.cpp
```diff
@@ -52,21 +51,20 @@ void ARRGameState::StartSim()
 {
     UE_LOG_WITH_INFO(LogRapyutaCore, Display, TEXT("[Start Sim with Num of SceneInstances: %d]"), SCENE_INSTANCES_NUM);
 
-    URRGameSingleton::Get()->PrintSimConfig();
```
Tadinu, line 55:
```
auto* gameSingleton = URRGameSingleton::Get();
if (gameSingleton)
{
    gameSingleton->PrintSimConfig();
}
```

Source/RapyutaSimulationPlugins/Private/Core/RRGameState.cpp
```diff
@@ -52,21 +51,20 @@ void ARRGameState::StartSim()
 {
     UE_LOG_WITH_INFO(LogRapyutaCore, Display, TEXT("[Start Sim with Num of SceneInstances: %d]"), SCENE_INSTANCES_NUM);
 
-    URRGameSingleton::Get()->PrintSimConfig();
     PrintSimConfig();
 
     GameMode = URRCoreUtils::GetGameMode<ARRGameMode>(this);
     check(GameMode);
```
Tadinu: Should rem check(GameMode) also?

Source/RapyutaSimulationPlugins/Private/Core/RRPlayerController.cpp
```diff
@@ -42,26 +35,31 @@ void ARRPlayerController::BeginPlay()
...
+#if RAPYUTA_USE_SCENE_DIRECTOR
     check(ActorCommon);
     check(ActorCommon->SceneInstanceId == SceneInstanceId);
 
     SceneCamera = ActorCommon->SceneCamera;
     check(SceneCamera);
-    if (GameMode->IsDataSynthSimType())
-    {
-        Possess(SceneCamera);
-    }
```
Tadinu: I suppose we should keep this since might be presumptuous to automatically `Possess(SceneCamera)`, which will lock user camera view,

Source/RapyutaSimulationPlugins/Private/Core/RRROS2GameMode.cpp
```diff
@@ -74,7 +74,11 @@ void ARRROS2GameMode::InitROS2()
     MainROS2Node = UROS2NodeComponent::CreateNewNode(this, MainROS2NodeName, TEXT("/"));
 
     // MainSimState
-    check(MainSimState);
+    if(MainSimState == nullptr)
+    {
+        UE_LOG_WITH_INFO(LogRapyutaCore, Error, TEXT("Failed to create ROS2Node."));
+        return;
+    }
```
Tadinu:
```suggestion
    if(MainSimState == nullptr)
    {
        UE_LOG_WITH_INFO(LogRapyutaCore, Error, TEXT("Failed to create MainSimState."));
        return;
    }
```

Source/RapyutaSimulationPlugins/Private/Robots/RRRobotROS2Interface.cpp
```diff
@@ -401,7 +403,13 @@ void URRRobotROS2Interface::JointStateCallback(const UROS2GenericMsg* Msg)
         AsyncTask(ENamedThreads::GameThread,
                   [this, joints, jointControlType]
                   {
-                      check(IsValid(Robot));
+                      if(!IsValid(Robot))
+                      {
+                        UE_LOG_WITH_INFO_NAMED(LogRapyutaCore,
+                                    Warning,
+                                    TEXT("Robot is nullptr. RobotROS2Interface::Robot must not be nullptr."));
```
Tadinu:
```suggestion
                        UE_LOG_WITH_INFO_NAMED(LogRapyutaCore,
                                    Error,
                                    TEXT("Robot is nullptr. RobotROS2Interface::Robot must not be nullptr."));
```

Source/RapyutaSimulationPlugins/Private/Sensors/RR2DLidarComponent.cpp
```diff
@@ -330,8 +330,10 @@ FROSLaserScan URR2DLidarComponent::GetROS2Data()
                                                               FVector::DotProduct(HitSurfaceNormal, -RayDirection),
                                  IntensityNonReflective,
                                  IntensityReflective);
-                check(Intensity >= IntensityNonReflective);
-                check(Intensity <= IntensityReflective);
+                if (Intensity <= IntensityNonReflective || Intensity <= IntensityReflective)
+                {
+                    UE_LOG_WITH_INFO(LogTemp, Warning, TEXT("Intensity is outof range. Something is wrong."));
+                }
```
Tadinu:
```suggestion
                if ((Intensity <= IntensityNonReflective) || (Intensity <= IntensityReflective))
                {
                    UE_LOG_WITH_INFO(LogRapyutaCore, Warning, TEXT("Intensity is outof range. Something is wrong."));
                }
```

Source/RapyutaSimulationPlugins/Private/Sensors/RR3DLidarComponent.cpp
```diff
@@ -333,8 +333,10 @@ FROSPointCloud2 URR3DLidarComponent::GetROS2Data()
                                                               FVector::DotProduct(HitSurfaceNormal, -RayDirection),
                                  IntensityNonReflective,
                                  IntensityReflective);
-                check(UnnormalizedIntensity >= IntensityNonReflective);
-                check(UnnormalizedIntensity <= IntensityReflective);
+                if (UnnormalizedIntensity <= IntensityNonReflective || UnnormalizedIntensity <= IntensityReflective)
+                {
+                    UE_LOG_WITH_INFO(LogTemp, Warning, TEXT("Normalized intensity is outof range. Something is wrong."));
+                }
```
Tadinu:
```suggestion
                if ((UnnormalizedIntensity <= IntensityNonReflective) || (UnnormalizedIntensity <= IntensityReflective))
                {
                    UE_LOG_WITH_INFO(LogRapyutaCore, Warning, TEXT("Normalized intensity is outof range. Something is wrong."));
                }
```

Source/RapyutaSimulationPlugins/Private/Tools/OccupancyMapGenerator.cpp
```diff
@@ -62,7 +62,7 @@ void AOccupancyMapGenerator::BeginPlay()
     }
 
     // write to file
-    check(WriteToFile(NCellsX, NCellsY, Origin.X / 100.f, -(Center.Y + Extent.Y) / 100.f));
+    WriteToFile(NCellsX, NCellsY, Origin.X / 100.f, -(Center.Y + Extent.Y) / 100.f);
```
Tadinu: Some error log upon failure?

Source/RapyutaSimulationPlugins/Private/Tools/RRROS2SkeletalMeshStatePublisher.cpp
```diff
@@ -35,7 +35,6 @@ void URRROS2SkeletalMeshStatePublisher::SetTargetRobot(ARobotVehicle* InRobot)
         SkeletalMeshComp = skeletalMeshComponents[0];
         const auto bonesNum = SkeletalMeshComp->GetBoneSpaceTransforms().Num();
         UE_LOG_WITH_INFO(LogRapyutaCore, Log, TEXT("[%s] has %d bones"), *SkeletalMeshComp->GetName(), bonesNum);
-        check(bonesNum > 0);
```
Tadinu, line 38: error log for (bonesNum == 0)

Source/RapyutaSimulationPlugins/Private/Core/RRGameMode.cpp
```diff
@@ -17,61 +17,63 @@
...
 
 void ARRGameMode::PrintSimConfig() const
 {
-    UE_LOG_WITH_INFO(LogRapyutaCore, Display, TEXT("SIM GAME MODE CONFIG -----------------------------"));
-    UE_LOG_WITH_INFO(LogRapyutaCore, Display, TEXT("- bBenchmark: %d"), bBenchmark);
+    UE_LOG_WITH_INFO(LogRapyutaCore, Verbose, TEXT("SIM GAME MODE CONFIG -----------------------------"));
+    UE_LOG_WITH_INFO(LogRapyutaCore, Verbose, TEXT("- bBenchmark: %d"), bBenchmark);
 }
 
 void ARRGameMode::PrintUEPreprocessors() const
```
Tadinu: This function as its name is meant to print useful logs about special UE-builtin settings generally. So if for some reason, it should be turned off by the caller upon certain condition, instead of `Verbose` them all.

Tadinu (approved):
Thanks for the de-coupling refactoring, I only had some minor comments.
Just a note that these would mostly affect DataSynth app only, which is not in dev for now so should be fine.

Source/RapyutaSimulationPlugins/Private/Tools/OccupancyMapGenerator.cpp
yuokamoto: Currently the function return always false

Source/RapyutaSimulationPlugins/Private/Core/RRGameInstance.cpp
yuokamoto: Do you think switch by `RAPYUTA_SIM_DEBUG`  is not convinent?

Source/RapyutaSimulationPlugins/Private/Core/RRGameState.cpp
yuokamoto: Do you needs to print sim config from GameState?
It seems GameSingleton is used only here in RRGameState.

Source/RapyutaSimulationPlugins/Public/Core/RRActorCommon.h
```diff
@@ -564,6 +565,7 @@ DECLARE_DELEGATE_TwoParams(FOnMeshActorFullyCreated, bool /* bCreationResult */,
 
 /**
  * @brief Scene instance's common object which houses Plugin-specific dynamic properties and implement objects-related API (Spawning, teleporting, etc.)
+ * If USE_SCENE_DIRECTOR is disabled, single ActorCommon is shared among all RRBaseActor.
```
Tadinu:
```suggestion
 * If USE_SCENE_DIRECTOR is disabled, there is only one single ActorCommon being shared among all RRBaseActor.
```

Source/RapyutaSimulationPlugins/Private/Core/RRBaseActor.cpp
yuokamoto: As far as I checked only GameState and GameSingleton are used.
I think it is better to remove other pointers until those become needed again to avoid confusion

Source/RapyutaSimulationPlugins/Private/Core/RRGameInstance.cpp
Tadinu: Yeah so RAPYUTA_SIM_DEBUG is actually just a dev-marker for non-ready source, thus not meant as a directive to be turned on/off globally. If we want to numb a group of certain logs, we should use a specific one like `RAPYUTA_SIM_INIT_DEBUG`

Source/RapyutaSimulationPlugins/Private/Tools/OccupancyMapGenerator.cpp
Tadinu: always [true](https://github.com/rapyuta-robotics/RapyutaSimulationPlugins/blob/devel/Source/RapyutaSimulationPlugins/Private/Tools/OccupancyMapGenerator.cpp#L90)? So may need to refactor it a bit to reflect the return of `FFileHelper::SaveArrayToFile()` I guess

Source/RapyutaSimulationPlugins/Private/Core/RRGameState.cpp
Tadinu: Yeah I could not find another place to print `URRGameSingleton`'s configs, which is actually initialized by UE itself. We just config its type in DefaultEngine.ini and fetch it during sim init, & it does not have an init function like StartSim()

Source/RapyutaSimulationPlugins/Private/Core/RRBaseActor.cpp
Tadinu: So the idea of `RapyutaSimPlugins` is to provide a suite of handy `RR~` actors that should come along but with minimal coupling. I suppose the confusion comes from the fact that we used `check` for debugging, which is now already removed now in this PR. In fact,
   * `GameMode` is very essential eg used [here](https://github.com/rapyuta-robotics/RapyutaSimulationPlugins/blob/devel/Source/RapyutaSimulationPlugins/Private/Core/RRProceduralMeshComponent.cpp#L122), as of `ROS2GameMode`, which contains SimState.
-> Having a convenient pointer here would facilitate us to quickly fetch `SimState` from any `RRBaseActor` like [here](https://github.com/rapyuta-robotics/RapyutaSimulationInternal/blob/RuntimePhysicsPlatter/Source/RapyutaSimInternal/Private/Forklift/RRPlatter.cpp#L940), without having to call the expensive `GetAllActorsOfClass` everytime we need to fetch it.
   * `PlayerController` is mostly used for DataSynth app.

Anyway, they are fetched here mostly for convenience to be used later, so we could just have to do a simple cast in client project/plugin without having to fetch them again by `GetWorld()->GetAuthGameMode()`

Tadinu: So the idea of `RapyutaSimPlugins` is to provide a suite of handy `RR~` actors that should come along but with minimal coupling. I suppose the confusion comes from the fact that we used `check` for debugging, which is now already removed in this PR. In fact,
   * `GameMode` is very essential eg used [here](https://github.com/rapyuta-robotics/RapyutaSimulationPlugins/blob/devel/Source/RapyutaSimulationPlugins/Private/Core/RRProceduralMeshComponent.cpp#L122), as of `ROS2GameMode`, which contains SimState.
-> Having a convenient pointer here would facilitate us to quickly fetch `SimState` from any `RRBaseActor` like [here](https://github.com/rapyuta-robotics/RapyutaSimulationInternal/blob/RuntimePhysicsPlatter/Source/RapyutaSimInternal/Private/Forklift/RRPlatter.cpp#L940), without having to call the expensive `GetAllActorsOfClass` everytime we need to fetch it.
   * `PlayerController` is mostly used for DataSynth app.

Anyway, they are fetched here mostly for convenience to be used later, so we could just have to do a simple cast in client project/plugin without having to fetch them again by `GetWorld()->GetAuthGameMode()`

Source/RapyutaSimulationPlugins/Private/Core/RRGameState.cpp
yuokamoto: Can't you use constructor?

Source/RapyutaSimulationPlugins/Private/Tools/RRROS2SkeletalMeshStatePublisher.cpp
yuokamoto: I don't think bonesNum==0 don't cause error

Source/RapyutaSimulationPlugins/Private/Core/RRGameInstance.cpp
yuokamoto: Added RAPYUTA_SIM_VERBOSE

Source/RapyutaSimulationPlugins/Private/Core/RRBaseActor.cpp
yuokamoto: OK then,
I changed variable name for clarification

Source/RapyutaSimulationPlugins/Private/Core/RRGameState.cpp
Tadinu: We cannot since in ctor, value assigning from INI does not happen yet

Source/RapyutaSimulationPlugins/Private/Tools/RRROS2SkeletalMeshStatePublisher.cpp
Tadinu: Yeap, a skeletal mesh component is definitely invalid upon no bones, but more essentially, it seems this class is unused?

Source/RapyutaSimulationPlugins/Private/Core/RRBaseActor.cpp
```diff
@@ -43,19 +42,26 @@ void ARRBaseActor::PreInitializeComponents()
...
+    // SETUP + CONFIGURE GAME & SIM COMMON OBJECTS
     // NOTE: These are used ONLY for dynamic runtime actors.
     // In child BP actors, except for GameSingleton, other Game framework entities maybe not available yet
-    GameMode = URRCoreUtils::GetGameMode<ARRGameMode>(this);
-    GameState = URRCoreUtils::GetGameState<ARRGameState>(this);
+#if RAPYUTA_USE_SCENE_DIRECTOR
+    RRGameState = URRCoreUtils::GetGameState<ARRGameState>(this);
+    check(RRGameState)
+#endif
+        RRROS2GameMode = URRCoreUtils::GetGameMode<ARRROS2GameMode>(this);
```
Tadinu: Since ARRGameMode is still used for `IsSimType()` or `IsDataSynthSimType()` in DataSynthApp, please keep
`RRGameMode = URRCoreUtils::GetGameMode<ARRGameMode>(this);` without check or verify

Source/RapyutaSimulationPlugins/Private/Tools/RRROS2SkeletalMeshStatePublisher.cpp
yuokamoto: yes this class is not used for now.

Source/RapyutaSimulationPlugins/Private/Core/RRBaseActor.cpp
```diff
@@ -43,19 +42,26 @@ void ARRBaseActor::PreInitializeComponents()
...
-    // SETUP + CONFIGURE ESSENTIAL GAME & SIM COMMON OBJECTS
+    // SETUP + CONFIGURE GAME & SIM COMMON OBJECTS
     // NOTE: These are used ONLY for dynamic runtime actors.
     // In child BP actors, except for GameSingleton, other Game framework entities maybe not available yet
-    GameMode = URRCoreUtils::GetGameMode<ARRGameMode>(this);
-    GameState = URRCoreUtils::GetGameState<ARRGameState>(this);
+#if RAPYUTA_USE_SCENE_DIRECTOR
+    RRGameState = URRCoreUtils::GetGameState<ARRGameState>(this);
+    check(RRGameState)
+#endif
```
Tadinu:
```suggestion
    RRGameState = URRCoreUtils::GetGameState<ARRGameState>(this);
```
for still being used [here](https://github.com/rapyuta-robotics/RapyutaSimulationInternal/blob/devel/Source/RapyutaSimInternal/Private/Warehouse/RRPallet.cpp#L20)
So basically all of them are still in use

yuokamoto: if those are used only in one child class. I think it is make sense to move it to that class instead of keeping it in base class.

Source/RapyutaSimulationPlugins/Private/Core/RRBaseActor.cpp
yuokamoto: Where I can find?

Tadinu: I mentioned it [here](https://github.com/rapyuta-robotics/RapyutaSimulationPlugins/pull/154#discussion_r1181502911)
Also a quick Ctrl+Shift+F will show it like [here](https://github.com/rapyuta-robotics/RapyutaSimulationPlugins/blob/devel/Source/RapyutaSimulationPlugins/Private/Core/RRProceduralMeshComponent.cpp#L122)

Source/RapyutaSimulationPlugins/Private/Core/RRBaseActor.cpp
Tadinu: So our RRGame~ actors could be used anywhere in our Sim, not particularly a mesh or non-mesh actor, so the essential idea is to avoid duplication in child classes, which is the very reason why we put it here in the base one.

Source/RapyutaSimulationPlugins/Private/Core/RRBaseActor.cpp
yuokamoto: Why do  you needs to call `GetAllActorsOfClass` to fetch SimState if there is no ptr? Can' you use GetGameMode(), cast 
 and access SimState work?

Tadinu: Yeap that should be the correct way to fetch it, though sometimes not always followed like [here](https://github.com/rapyuta-robotics/RapyutaSimulationInternal/blob/devel/Source/RapyutaSimInternal/Private/Human/RRHumanROS2Interface.cpp#L126).
Essentially, repeated `GetGame~()` call every time it needs to be used should be avoided for reasons:
* It is a UE-api call, so we don't have full control over the overheads it takes. (like it typically involves `GetWorld()` which is not a trivial call at all).
* Even if we know it's cheap to call, it's always cheaper to fetch it once only (except it is changed at runtime but Game~ actors do not) and use during Sim run, which is a common practice also applied to fetching UE-asset resource.

yuokamoto: - Why you didn't follow correct way there?
- Are there any documentation that shows `GetWorld()` is not a trivial call? I think it is difficult to know all api which is slow but we can have list of those function.

Source/RapyutaSimulationPlugins/Private/Core/RRBaseActor.cpp
yuokamoto: I do not fully agreed but reverted since that is not main topic of this pr.

Tadinu: Yp, let's do it step by step. My thinking is any refactoring should come with valid specific reasons, otherwise, it might raise some questions to future maintainers on the consistency of our large-growing source base.
* This PR is meant to reduce the coupling with RRGame~ meaning `check` should not be imposed, meaning the use of `RRGame~` is optional (as up to users to decide it should be used or not).
* There was also no mention of the rationale in the PR description that member pointers should be also removed, which would cause an impact to client source.

Source/RapyutaSimulationPlugins/Private/Core/RRBaseActor.cpp
Tadinu: Yeah,
* It is not my code, it is my observation of the inefficient use of the API, which happens when the class does not hold a quick pointer to `RRGame~`.
* If you look at UE-builtin `GetWorld()/GetGame~()` implementation, it does a couple of queries and lookups, which might be or not be much of overheads but apparently, it's always cheaper if we already have a pointer `UWorld*` (of course as long as it is not updated during Sim run). That practice is applied to any type of pointer referencing in general, as the less referencing is done, less lookup work for the processor. 
* Anyway, since they are UE-apis thus better not to use them too expediently since their implementation are prone to be updated, which is also why we have Utils functions instead of direct calls to UE-apis for:
+ To reduce duplication.
+ To reduce coupling like if they are deprecated, we just need to update Util function instead of going to all places in oursourcebase to update it.

PR #158, opened 2023-05-07
Add runtime UI/RRUserWidget used as RRBaseRobot's tooltip

Tadinu:
* Add `URRUserWidget` inheriting from `UUserWidget`, containing a default TextBlock
* `ARRBaseRobot` add `bTooltipEnabled` (defaulted as False) + tooltip/label component as `URRUserWidget`, used in case of showing its debug content like id name, status, etc.


Source/RapyutaSimulationPlugins/Public/Robots/RRBaseRobot.h
```diff
@@ -443,6 +450,27 @@ class RAPYUTASIMULATIONPLUGINS_API ARRBaseRobot : public ARRBaseActor
     UFUNCTION(BlueprintImplementableEvent, BlueprintCallable)
     void BPConfigureMovementComponent();
 
+    // TOOLTIP --
+    UPROPERTY()
+    uint8 bTooltipEnabled : 1;
+    //! Tooltip widget component
+    UPROPERTY()
+    TObjectPtr<UWidgetComponent> TooltipComp = nullptr;
```
yuokamoto: it maybe better to rename to general name since it can be used for any widget rather than tooltip

yuokamoto (commented):
nice to have option to show/hide at runtime since this is useful not only for debug but also for normal usecase

yuokamoto (approved):

PR #159, opened 2023-05-07
Utils for runtime physics robot

Tadinu:
`URRStaticMeshComponent` runtime collision
* add `bUseSimpleCollision` dft true, `bUseComplexCollision` dft false
* add `CreateMesh()`, creating an either visual or collision mesh, invoked by `CreateMeshBody()`
* Use VHACD to auto-generate runtime mesh simple collision info
`URRUObjectUtils` add SetActorCollisionEnabled()
`URRBaseRobot` StopMovement()
- Call MoveComp->StopMovementImmediately()
- RRFloatMoveComp reimpl StopMovementImmediately() resetting `AngularVelocity`
`URRURDFParser` Remove incorrect check of `IsPlainWheeledVehicleModel()` as deciding which links are wheels


Source/RapyutaSimulationPlugins/Public/Core/RRUObjectUtils.h
```diff
@@ -591,6 +591,15 @@ class RAPYUTASIMULATIONPLUGINS_API URRUObjectUtils : public UBlueprintFunctionLi
         return GetRelativeQuatFrom(InActor, InBaseActor).Rotator();
     }
 
+    FORCEINLINE static void SetActorCollisionEnabled(AActor* InActor, bool bInCollisionOn)
```
yuokamoto, line 599: please add doxygen comment.

Source/RapyutaSimulationPlugins/Public/Robots/RRRobotStructs.h
```diff
@@ -332,7 +332,7 @@ struct RAPYUTASIMULATIONPLUGINS_API FRRRobotJointProperty
                 return ACM_Locked;
 
             case ERRRobotJointType::REVOLUTE:
-                return FMath::IsNearlyZero(UpperLimit) ? ACM_Locked : (UpperLimit >= M_PI) ? ACM_Free : ACM_Limited;
+                return (UpperLimit >= M_PI) ? ACM_Free : ACM_Limited;
```
yuokamoto: This is not always true. some robot joints rotational limit is +- 360
https://www.universal-robots.com/media/50895/ur10_en.pdf

Source/RapyutaSimulationPlugins/RapyutaSimulationPlugins.Build.cs
```diff
@@ -44,7 +44,7 @@ public RapyutaSimulationPlugins(ReadOnlyTargetRules Target) : base(Target)
         bEnableExceptions = true;
 
         PublicDependencyModuleNames.AddRange(new string[] { "Core", "CoreUObject", "Engine", "InputCore", "ImageWrapper", "RenderCore", "Renderer", "RHI", "PhysicsCore", "XmlParser", "IESFile",
-                                                            "AIModule", "NavigationSystem", "TimeManagement", "Json",
+                                                            "AIModule", "NavigationSystem", "TimeManagement", "Json", "PhysicsUtilities",
```
yuokamoto: just comment
we maybe better to sort plugins later here

Source/RapyutaSimulationPlugins/Private/Core/RRStaticMeshComponent.cpp
```diff
@@ -203,59 +208,111 @@ UStaticMesh* URRStaticMeshComponent::CreateMeshBody(const FRRMeshData& InMeshDat
...
+    // Generate complex in case of no simple collision
+    if (false == bUseDefaultSimpleCollision)
+    {
+        auto* bodySetup = visualMesh->GetBodySetup();
+        if (bUseComplexCollision)
+        {
+            visualMesh->ComplexCollisionMesh = CreateMesh(InMeshData, false);
+            bodySetup->CollisionTraceFlag = ECollisionTraceFlag::CTF_UseComplexAsSimple;
+        }
+        else
```
yuokamoto, line 295: Is this elase{} non default simple collision? When we use this?

Source/RapyutaSimulationPlugins/RapyutaSimulationPlugins.Build.cs
Tadinu: Ideally modules should be grouped into their functionality relevancy (like Core, Physics, AI, Nav, etc.) for easier management & Our plugins should be put at the end

Source/RapyutaSimulationPlugins/Private/Core/RRStaticMeshComponent.cpp
Tadinu: Yes all of those are wrapped inside `if(!bUseDefaultSimpleCollision)`
Basically, Complex collision, as using an external mesh, only works for static objects in UE since moveable one is expensive to simulate
https://docs.unrealengine.com/5.1/en-US/simple-versus-complex-collision-in-unreal-engine
Custom simple collision, as generated by VHACD, works for both static/movable objects.

yuokamoto (approved):

PR #160, opened 2023-05-11
RRBaseRobot AngularVel in Degree

Tadinu:
* ROS twist msg's angular is in `rad` as SI unit https://www.ros.org/reps/rep-0103.html#units
* UE angular vel has been in `deg` to be consistent with `FRotator`, which is in `deg`
-> Conversion is done [here](https://github.com/rapyuta-robotics/RapyutaSimulationPlugins/blob/devel/Source/RapyutaSimulationPlugins/Private/Robots/RRRobotROS2Interface.cpp#L301) as receiving Twist msg
This PRs set `ARRBaseRobot/URobotVehicleMovementComponent`'s AngularVelocity to Degree for clarity in its usage

NOTE:
In DifferentialDriveComponent/RobotVehicleMovementComponent's UpdateMovement(), which is called per tick, `DegreesToRadians(AngularVel.Yaw)` is now needed for the calculation.
-> This could be avoided by using a local member `AngularVelRad`, however it could also cause a bit more overheads on maintenance, which is left to be considered.

### DIFF
--- a/Source/RapyutaSimulationPlugins/Private/Core/RRNetworkPlayerController.cpp
+++ b/Source/RapyutaSimulationPlugins/Private/Core/RRNetworkPlayerController.cpp
@@ -215,7 +215,7 @@ void ARRNetworkPlayerController::ServerSetLinearVel_Implementation(ARRBaseRobot*
 void ARRNetworkPlayerController::ServerSetAngularVel_Implementation(ARRBaseRobot* InServerRobot,
                                                                     float InClientTimeStamp,
                                                                     const FRotator& InClientRobotRotation,
-                                                                    const FRotator& InAngularVel)
+                                                                    const FVector& InAngularVel)
 {
 #if RAPYUTA_SIM_DEBUG
     UE_LOG_WITH_INFO_NAMED(
@@ -225,7 +225,8 @@ void ARRNetworkPlayerController::ServerSetAngularVel_Implementation(ARRBaseRobot
     if (robot)
     {
         float serverCurrentTime = UGameplayStatics::GetRealTimeSeconds(GetWorld());
-        robot->SetActorRotation(InClientRobotRotation + InAngularVel * (serverCurrentTime - InClientTimeStamp));
+        robot->SetActorRotation(InClientRobotRotation +
+                                FRotator::MakeFromEuler(InAngularVel) * (serverCurrentTime - InClientTimeStamp));
         //NOTE: Don't use ARRBaseRobot::SetAngularVel() here, which is only for client
         robot->TargetAngularVel = InAngularVel;
     }
--- a/Source/RapyutaSimulationPlugins/Private/Core/RRNetworkPlayerController.cpp
+++ b/Source/RapyutaSimulationPlugins/Private/Core/RRNetworkPlayerController.cpp
@@ -17,7 +17,6 @@
 #include "Core/RRCoreUtils.h"
 #include "Core/RRUObjectUtils.h"
 #include "Robots/RRBaseRobot.h"
-#include "Robots/RRBaseRobot.h"
 #include "Robots/RRRobotROS2Interface.h"
 #include "Tools/RRROS2SimulationStateClient.h"
 #include "Tools/SimulationState.h"
@@ -216,7 +215,7 @@ void ARRNetworkPlayerController::ServerSetLinearVel_Implementation(ARRBaseRobot*
 void ARRNetworkPlayerController::ServerSetAngularVel_Implementation(ARRBaseRobot* InServerRobot,
                                                                     float InClientTimeStamp,
                                                                     const FRotator& InClientRobotRotation,
-                                                                    const FVector& InAngularVel)
+                                                                    const FRotator& InAngularVel)
 {
 #if RAPYUTA_SIM_DEBUG
     UE_LOG_WITH_INFO_NAMED(
@@ -226,7 +225,7 @@ void ARRNetworkPlayerController::ServerSetAngularVel_Implementation(ARRBaseRobot
     if (robot)
     {
         float serverCurrentTime = UGameplayStatics::GetRealTimeSeconds(GetWorld());
-        robot->SetActorRotation(InClientRobotRotation + InAngularVel.Rotation() * (serverCurrentTime - InClientTimeStamp));
+        robot->SetActorRotation(InClientRobotRotation + InAngularVel * (serverCurrentTime - InClientTimeStamp));
         //NOTE: Don't use ARRBaseRobot::SetAngularVel() here, which is only for client
         robot->TargetAngularVel = InAngularVel;
     }
--- a/Source/RapyutaSimulationPlugins/Private/Robots/RRRobotROS2Interface.cpp
+++ b/Source/RapyutaSimulationPlugins/Private/Robots/RRRobotROS2Interface.cpp
@@ -298,7 +298,7 @@ void URRRobotROS2Interface::MovementCallback(const UROS2GenericMsg* Msg)
         FROSTwist twist;
         twistMsg->GetMsg(twist);
         const FVector linear(URRConversionUtils::VectorROSToUE(twist.Linear));
-        const FVector angular(FMath::RadiansToDegrees(URRConversionUtils::RotationROSToUE(twist.Angular)));
+        const FRotator angular(URRConversionUtils::RotationROSToUE(twist.Angular));
 
         // (Note) In this callback, which could be invoked from a ROS working thread,
         // thus any direct referencing to its member in this GameThread lambda needs to be verified.
--- a/Source/RapyutaSimulationPlugins/Private/Robots/RRRobotROS2Interface.cpp
+++ b/Source/RapyutaSimulationPlugins/Private/Robots/RRRobotROS2Interface.cpp
@@ -297,8 +297,8 @@ void URRRobotROS2Interface::MovementCallback(const UROS2GenericMsg* Msg)
         // probably should not stay in msg though
         FROSTwist twist;
         twistMsg->GetMsg(twist);
-        const FVector linear(URRConversionUtils::VectorROSToUE(twist.Linear));
-        const FRotator angular(URRConversionUtils::RotationROSToUE(twist.Angular));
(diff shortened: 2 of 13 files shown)


Source/RapyutaSimulationPlugins/Public/Core/RRConversionUtils.h
```diff
@@ -152,15 +152,29 @@ class URRConversionUtils : public UBlueprintFunctionLibrary
...
     }
 
+    /**
+     * @brief Convert ROS Rotation Euler (rad) to UE Vector retaining unit as [rad]
+     * @param Input
+     * @return FVector
+     */
     UFUNCTION(BlueprintCallable, Category = "Conversion")
-    static FVector RotationROSToUE(const FVector& Input)
+    static FVector RotationROSToUEVector(const FVector& Input)
```
yuokamoto: Based on the naming convention, i.e. ROSToUE does orieantation and unit conversion, it is better to rad to degree here

yuokamoto:
- I don't think using Rotator is making sense since Rotator is designed to limit the value to [0, 360]. 
https://docs.unrealengine.com/4.27/en-US/BuildingWorlds/LevelEditor/Details/Properties/VectorAndRotator/
- UE API also uses angular velocity as FVector https://docs.unrealengine.com/4.27/en-US/API/Runtime/Engine/Components/UPrimitiveComponent/GetPhysicsAngularVelocity/

Source/RapyutaSimulationPlugins/Public/Core/RRConversionUtils.h
Tadinu: So we actually use it also for Odom msg conversion [here](https://github.com/rapyuta-robotics/RapyutaSimulationPlugins/blob/devel/Source/RapyutaSimulationPlugins/Public/Core/RRConversionUtils.h#L117), where no rad<->deg should happen.
I added a param for conversion up to the caller.

Tadinu:
ah right I forgot about it, though it's not done automatically in FRotator ctor but its operation like `Equals` does normalize, so I reverted back to Vector.

Source/RapyutaSimulationPlugins/Private/Core/RRNetworkPlayerController.cpp
```diff
@@ -226,7 +225,8 @@ void ARRNetworkPlayerController::ServerSetAngularVel_Implementation(ARRBaseRobot
     if (robot)
     {
         float serverCurrentTime = UGameplayStatics::GetRealTimeSeconds(GetWorld());
-        robot->SetActorRotation(InClientRobotRotation + InAngularVel.Rotation() * (serverCurrentTime - InClientTimeStamp));
+        robot->SetActorRotation(InClientRobotRotation +
```
yuokamoto, line 228: what is the difference between 
InAngularVel.Rotation() and FRotator::MakeFromEuler(InAngularVel)

Source/RapyutaSimulationPlugins/Private/Robots/RRRobotROS2Interface.cpp
```diff
@@ -297,8 +297,8 @@ void URRRobotROS2Interface::MovementCallback(const UROS2GenericMsg* Msg)
         // probably should not stay in msg though
         FROSTwist twist;
         twistMsg->GetMsg(twist);
```
yuokamoto, line 299: what is the difference between these two initialization?

Source/RapyutaSimulationPlugins/Public/Drives/DifferentialDriveComponent.h
```diff
@@ -104,13 +105,17 @@ class RAPYUTASIMULATIONPLUGINS_API UDifferentialDriveComponent : public URobotVe
     float MaxForce = 1000.f;
 
 protected:
+    //! [cm]
     UPROPERTY()
     float WheelPerimeter = 6.28f;
 
+    //! [rad]
```
yuokamoto: cm

Source/RapyutaSimulationPlugins/Public/Drives/DifferentialDriveComponent.h
```diff
@@ -104,13 +105,17 @@ class RAPYUTASIMULATIONPLUGINS_API UDifferentialDriveComponent : public URobotVe
...
 
 protected:
+    //! [cm]
     UPROPERTY()
     float WheelPerimeter = 6.28f;
 
+    //! [rad]
     UPROPERTY()
     float PoseEncoderX = 0.f;
+    //! [rad]
```
yuokamoto: cm

Source/RapyutaSimulationPlugins/Public/Drives/DifferentialDriveComponent.h
```diff
@@ -104,13 +105,17 @@ class RAPYUTASIMULATIONPLUGINS_API UDifferentialDriveComponent : public URobotVe
...
     UPROPERTY()
     float PoseEncoderY = 0.f;
+    //! [rad]
```
yuokamoto, line 118: degree

Source/RapyutaSimulationPlugins/Public/Core/RRConversionUtils.h
yuokamoto: not 100% sure but it seems bug and OdomROSToUE should return rad since OdomData in ROS is radian.

yuokamoto:
Thanks for your PR. it seems there are many places that uses degree mistakenly instead of radian. hope this will improve navigation behaviour.(I don't know why nav kind of work with those bug... feedback control maybe robust enough)

Source/RapyutaSimulationPlugins/Private/Core/RRNetworkPlayerController.cpp
Tadinu: So `InAngularVel.Rotation()` calls into [ToOrientationRotator(),](https://github.com/EpicGames/UnrealEngine/blob/release/Engine/Source/Runtime/Core/Private/Math/UnrealMath.cpp#LL199C45-L199C65) but here we would just want the delta rotation I suppose.
Might you help confirm? @larryng-rapyuta

Source/RapyutaSimulationPlugins/Private/Robots/RRRobotROS2Interface.cpp
Tadinu: No difference actually, both calling the copy constructor, it's just easier to read with `=`
https://www.geeksforgeeks.org/copy-constructor-vs-assignment-operator-in-c/#:~:text=copy%20constructor%2C-,same,-as%20Test%20t3

Source/RapyutaSimulationPlugins/Public/Drives/DifferentialDriveComponent.h
Tadinu: As in the comment here, it is meant to be used directly for `cos()/sin()` & `FQuat` so should be rad?
https://github.com/rapyuta-robotics/RapyutaSimulationPlugins/blob/devel/Source/RapyutaSimulationPlugins/Private/Drives/DifferentialDriveComponent.cpp#L113
https://github.com/rapyuta-robotics/RapyutaSimulationPlugins/blob/devel/Source/RapyutaSimulationPlugins/Private/Drives/DifferentialDriveComponent.cpp#L124

Tadinu:
Ah previously, we didn't convert rad->deg as in `URRRobotROS2Interface::MovementCallback()` and thus use `AngularVel` as in Rad directly to calculate differential vels.
Recently we have that angular interpolation with angular vel set directly in deg, so it would cause impact if used together in normal mode I suppose.

yuokamoto:
> Ah previously, we didn't convert rad->deg as in `URRRobotROS2Interface::MovementCallback()` and thus use `AngularVel` as in Rad directly to calculate differential vels. Recently we have that angular interpolation with angular vel set directly in deg, so it would cause impact if used together in normal mode I suppose.

it was converted here https://github.com/rapyuta-robotics/RapyutaSimulationPlugins/blob/fc446e3a2131f20480c5735ec0d3da9c045427b5/Source/RapyutaSimulationPlugins/Private/Robots/RRRobotROS2Interface.cpp#L301

Tadinu:
Yeap the conversion was actually just added in the previous PR

Source/RapyutaSimulationPlugins/Public/Drives/DifferentialDriveComponent.h
yuokamoto: hmm I see. I thought it is better to variable in cm and deg as much as possible but in this case we can keep this rad. it maybe better to rename as PoseEncoderThetaRad to clarify

yuokamoto (approved):

larryng-rapyuta (approved):

Source/RapyutaSimulationPlugins/Public/Core/RRConversionUtils.h
```diff
@@ -119,13 +119,14 @@ class URRConversionUtils : public UBlueprintFunctionLibrary
         Output.Pose.Pose.Orientation = QuatUEToROS(Output.Pose.Pose.Orientation);
 
         Output.Twist.Twist.Linear = VectorUEToROS(Output.Twist.Twist.Linear);
-        Output.Twist.Twist.Angular = RotationUEToROS(Output.Twist.Twist.Angular, false);
+        Output.Twist.Twist.Angular = RotationUEVectorToROS(Output.Twist.Twist.Angular);
```
Tadinu, line 122: @yuokamoto  `FROSTwist` looks still in SI unit, and also is `FROSOdom` for consistency probably, since we don't do conversion here.
https://github.com/rapyuta-robotics/rclUE/blob/devel/Source/rclUE/Public/Msgs/ROS2Twist.h#L37
So should be no conversion here I suppose.

Tadinu, line 122: @yuokamoto Might need to update also `Angular` here to `degree`
https://github.com/rapyuta-robotics/RapyutaSimulationPlugins/blob/devel/Source/RapyutaSimulationPlugins/Private/Sensors/RRBaseOdomComponent.cpp#L127

PR #161, opened 2023-05-22
Add object saving to uasset on disk

Tadinu:
* `URRAssetUtils`: Add SaveObjectToAsset()/SaveObjectToAssetInModule()
* `URRGameSingleton`
   - InitializeResources() Also load skeletal objects (mesh, skeleton, physics assets) from uasset files
   - Made `RequestResourceLoading()` template to apply for generic resource data type
* `URRStaticMeshComponent::CreateMeshBody()` save generated static mesh to uasset to avoid dynamic creation later

### DIFF
--- a/Source/RapyutaSimulationPlugins/Public/Core/RRAssetUtils.h
+++ b/Source/RapyutaSimulationPlugins/Public/Core/RRAssetUtils.h
@@ -18,6 +18,7 @@
 
 #if WITH_EDITOR
 #include "AssetToolsModule.h"
+#include "IAssetTools.h"
 #endif
 
 // RapyutaSim
--- a/Source/RapyutaSimulationPlugins/Public/Core/RRAssetUtils.h
+++ b/Source/RapyutaSimulationPlugins/Public/Core/RRAssetUtils.h
@@ -13,6 +13,8 @@
 #include "Kismet/BlueprintFunctionLibrary.h"
 #include "Materials/Material.h"
 #include "PhysicalMaterials/PhysicalMaterial.h"
+#include "UObject/Package.h"
+#include "UObject/SavePackage.h"
 
 #if WITH_EDITOR
 #include "AssetToolsModule.h"
@@ -22,6 +24,7 @@
 #endif
 
 // RapyutaSim
+#include "Core/RRObjectCommon.h"
 #include "RapyutaSimulationPlugins.h"
 
 #include "RRAssetUtils.generated.h"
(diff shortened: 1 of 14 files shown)


Source/RapyutaSimulationPlugins/Public/Core/RRGameSingleton.h
```diff
@@ -26,6 +27,28 @@
 
 #include "RRGameSingleton.generated.h"
 
+template<const ERRResourceDataType InDataType>
```
yuokamoto, line 30: is this a standard way to define typenamne?

Tadinu: I suppose so, `using` as introduced since c++11 is favored over `typedef`
https://www.geeksforgeeks.org/cpp-using-vstypedef
https://www.nextptr.com/tutorial/ta1193988140/how-cplusplus-using-or-aliasdeclaration-is-better-than-typedef

Tadinu: Yeap, `InDataType` is non-type template parameter https://en.cppreference.com/w/cpp/language/template_parameters#:~:text=a%20non%2Dtype%20template%20parameter%3B
Besides, `using` as introduced since c++11 is favored over `typedef`
https://www.geeksforgeeks.org/cpp-using-vstypedef
https://www.nextptr.com/tutorial/ta1193988140/how-cplusplus-using-or-aliasdeclaration-is-better-than-typedef

Source/RapyutaSimulationPlugins/Public/Core/RRGameSingleton.h
```diff
@@ -103,30 +97,100 @@ class RAPYUTASIMULATIONPLUGINS_API URRGameSingleton : public UObject
...
     static constexpr const TCHAR* ASSETS_PROJECT_MODULE_NAME = TEXT("Game/RapyutaContents");
-    
+
     /**
-     * @brief Get the Assets Base Path object. 
+     * @brief Get name prefix of asset
+     * @param InDataType
+     * @sa [Asset naming convention](https://docs.unrealengine.com/5.2/en-US/recommended-asset-naming-conventions-in-unreal-engine-projects)
+     */
+    FORCEINLINE static constexpr const TCHAR* GetAssetNamePrefix(const ERRResourceDataType InDataType)
```
chalice-graeme, line 109: We should document in the README that these conventions are required.

Source/RapyutaSimulationPlugins/Public/Core/RRAssetUtils.h
```diff
@@ -343,20 +387,18 @@ class RAPYUTASIMULATIONPLUGINS_API URRAssetUtils : public UBlueprintFunctionLibr
...
 
-        // 2- Create package for the class asset
-        FString packageName = FString::Printf(TEXT("/Game/Blueprints/%s"), *InClassName.ToString());
-        FString assetName;
-        assetToolsModule.Get().CreateUniqueAssetName(packageName, TEXT(""), packageName, assetName);
-        UPackage* package = CreatePackage(*packageName);
-        check(package);
+        // 2- Create blueprint package for the class asset
+        const FString bpClassName = FString::Printf(TEXT("BP_%s"), *InClassName.ToString());
+        FString bpPackageName = FString::Printf(TEXT("/Game/Blueprints/%s"), *bpClassName);
```
chalice-graeme: It might be better to pass the package path as a parameter, and then allow it to be configured elsewhere by the user.

chalice-graeme (approved):

Source/RapyutaSimulationPlugins/Public/Core/RRGameSingleton.h
Tadinu: Yeah I added note to doxygen

yuokamoto (approved):
You need to add ue_asset at the btn of index.rst
you can check build result here
https://rapyutasimulationplugins.readthedocs.io/en/save_object_2_asset/

PR #168, opened 2023-06-01
Enable runtime static mesh saving to asset

Tadinu:
* `URRStaticMeshComponent` fix runtime mesh saving to uasset by turning on `meshDescParams.bCommitMeshDescription`, to  validated its mesh description
* `URRGameSingleton` add `GetDynamicAssetPath()`
* `URRAssetUtils` `SaveObjectToAssetInModule()/SaveObjectToAsset() `add `bSaveDuplicatedObject` param, which is required for skeletal suite objects, due to their inter-referencing, but not for standalone objects like static mesh


Content/Maps/entry_screen_widget.uasset
yuokamoto, line 1: What is the changes for these .uasset s?

yuokamoto (approved):

Content/Maps/entry_screen_widget.uasset
Tadinu: Noted in the commit comment due to the warning. Looks like we have to resave assets each time upgrading to new UE version

Tadinu: As noted in the commit comment due to the warning. Looks like we have to resave assets each time upgrading to new UE version

PR #169, opened 2023-06-07
Add BP Robot model data table loading

Tadinu:
RRRobotStructs Add `FRRRobotModelData`, used by `FRRRobotModelInfo` (dynamic runtime robot models info as loaded from URDF/SDF) & `FRRRobotModelTableRow` (static BP robot models info as stored in UE data table asset)
`ERRResourceDataType` add `UE_DATA_TABLE,` auto early loading existing data table dynamic assets

### DIFF
--- a/Source/RapyutaSimulationPlugins/Public/Robots/RRRobotStructs.h
+++ b/Source/RapyutaSimulationPlugins/Public/Robots/RRRobotStructs.h
@@ -31,7 +31,8 @@ enum class ERRRobotDescriptionType : uint8
 {
     NONE,
     URDF,
-    SDF
+    SDF,
+    UASSET,
 };
 
 /**
@@ -41,7 +42,7 @@ enum class ERRRobotDescriptionType : uint8
 UENUM()
 enum class ERRRobotStatus : uint8
 {
-    INVALID,             //! Invalid type, or not loaded successfully
+    INVALID,             //! Invalid status, eg not loaded successfully
     CREATING,            //! Being built up (loading the meshses, building the structure, etc.)
     IDLE,                //! Idling, ready for command order
     ORDER_PROCESSING,    //! Received command and in middle of processing
@@ -83,7 +84,7 @@ enum class ERRLidarSensorType : uint8
 UENUM()
 enum class ERRRobotJointType : uint8
 {
-    INVALID,
+    NONE,
     FIXED,
     FLOATING,
     PRISMATIC,
@@ -107,7 +108,7 @@ enum class ERRRobotJointStatus : uint8
 UENUM()
 enum class ERRRobotMeshComponentType : uint8
 {
-    INVALID,
+    NONE,
     STATIC_MESH,
     SKELETAL_MESH,
     PROCEDURAL_MESH,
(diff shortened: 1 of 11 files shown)


Source/RapyutaSimulationPlugins/Public/Robots/RRRobotStructs.h
```diff
@@ -1423,12 +1463,209 @@ struct RAPYUTASIMULATIONPLUGINS_API FRRRobotModelInfo : public TSharedFromThis<F
...
-        for (const auto& modelInfo : ChildModelsInfo)
+        for (const auto& modelInfo : ChildModelsData)
         {
             UE_LOG_WITH_INFO(LogTemp, Warning, TEXT("SUBMODEL"));
             modelInfo.PrintSelf();
         }
     }
+};    // END FRRRobotModelData
+
+USTRUCT(BlueprintType)
```
yuokamoto, line 1501: What is the main motivation to introduce these struct with almost just accessor?

Tadinu: As noted in the description, basically we need `FRRRobotModelInfo` as the ultimate in-memory robot model data, either loaded from URDF or from UDataTable.

They are meant for different purposes and thus different parent structs:
* `FRRRobotModelInfo : public TSharedFromThis<FRRRobotModelInfo, ESPMode::ThreadSafe>` -> to allow access as `TSharedRef` & thread-safe, since robot model info could be accessed from any thread.
* `FRRRobotModelTableRow : public FTableRowBase` -> only meant to store data in data table for BP robot users.

Both of these thus need to wrap the core robot model data stored in `FRRRobotModelData`.

Tadinu: As noted in the description, basically we need `FRRRobotModelInfo` as the ultimate in-memory robot model data, either loaded from URDF or from UDataTable.

They are meant for different purposes and thus different parent structs:
* `FRRRobotModelInfo : public TSharedFromThis<FRRRobotModelInfo, ESPMode::ThreadSafe>` -> to allow access as `TSharedRef` & thread-safe, since robot model info could be accessed from any thread. This does not support USTRUCT() thus cannot be used directly for storing data in UDataTable.
* `FRRRobotModelTableRow : public FTableRowBase` -> only meant to store data in data table for BP robot users.

Both of these thus need to wrap the core robot model data stored in `FRRRobotModelData`.

yuokamoto: For my understainding
- FRRRobotModelInfo provide Thread safe getter for FRRRobotModelData
- Does UDataTalbe require child struct of FTableRowBase only with data?

It is helpful to add doc/reference

yuokamoto (approved):

Source/RapyutaSimulationPlugins/Public/Robots/RRRobotStructs.h
Tadinu: Yeah it does
https://github.com/EpicGames/UnrealEngine/blob/release/Engine/Source/Runtime/Engine/Classes/Engine/DataTable.h#L315
As the name `FRRRobotModelTableRow` itself should already denote its meaning, I'll add some docs

PR #170, opened 2023-06-12
Asset saving: handle already existing name-sake

Tadinu:
`URRAssetUtils`
   * add `CreatePackageForSavingToAsset()`, called by `SaveObjectToAsset()` to allow creation of an independent package, which then could wrap some target objects before being saved to asset
   * `SaveObjectToAsset()` handle in case of already existing same-name asset

`URRBaseRobot` add `GetDynamicResourceAssetPath()`, retuning the saved path of a dynamically loaded resource asset
`URRConversionUtils` add `ToString() & ArrayToString()`

### DIFF
--- a/Source/RapyutaSimulationPlugins/Private/Core/RRAssetUtils.cpp
+++ b/Source/RapyutaSimulationPlugins/Private/Core/RRAssetUtils.cpp
@@ -179,19 +179,25 @@ UObject* URRAssetUtils::SaveObjectToAssetInModule(UObject* InObject,
                                                   const FString& InAssetUniqueName,
                                                   const TCHAR* InModuleName,
                                                   bool bSaveDuplicatedObject,
-                                                  bool bInStripEditorOnlyContent)
+                                                  bool bInStripEditorOnlyContent,
+                                                  bool bInAsyncSave,
+                                                  bool bInAlwaysOverwrite)
 {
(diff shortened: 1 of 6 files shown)


Source/RapyutaSimulationPlugins/Private/Core/RRAssetUtils.cpp
```diff
@@ -249,22 +276,31 @@ bool URRAssetUtils::SavePackageToAsset(UPackage* InPackage, UObject* InObject)
...
-                   *packageFileName);
+            if (bAsyncSave)
+            {
+                URRThreadUtils::DoAsyncTaskInThread<void>(
+                    []() { UPackage::WaitForAsyncFileWrites(); },
+                    [packageFileName]() { UE_LOG(LogRapyutaCore, Log, TEXT("SAVED UASSET %s"), *packageFileName); });
+            }
+            else
+            {
+                UE_LOG(LogRapyutaCore,
```
yuokamoto: UE_LOG_WITH_INFO

yuokamoto (commented):
Is it possible to overwrite existing asset?

Tadinu:
Yeah actually, `UPackage::Save()` always overwrite, I so just added option for it

yuokamoto (approved):

PR #177, opened 2023-06-19
Accomodate Cmdlet

Tadinu:
* `URRCoreUtils` add `IsRunningSimCommandlet<T>()` + `GetEditorWorld()`, used as running in cmdlet that does not have a default world
   * `ExecuteConsoleCommand() + GetActorCommon()` add treatment as running in cmdlet
* URRGameSingleton print configs, showing dynamic assets paths
   * add `GetDynamicBPAssetPath()`
   * `InitResources()` add `bInitRequestResourceLoading` dft true, allowing users to init the resource map only without having always to early load assets if not using
* RRRobotStructs add link/joint prop fetch apis
* Make sure Asset saving is Editor-only
* `URRAssetUtils add` GetAssetFilePathOnDisk()
* `ARRBaseRobot::InitPropertiesFromJSON()` return bool

### DIFF
--- a/Source/RapyutaSimulationPlugins/Private/Core/RRGameSingleton.cpp
+++ b/Source/RapyutaSimulationPlugins/Private/Core/RRGameSingleton.cpp
@@ -43,11 +43,13 @@ void URRGameSingleton::PrintSimConfig() const
     for (auto i = static_cast<int8>(ERRResourceDataType::NONE); i < static_cast<int8>(ERRResourceDataType::TOTAL); ++i)
     {
         const ERRResourceDataType dataType = static_cast<ERRResourceDataType>(i);
-        UE_LOG(
-            LogRapyutaCore, Display, TEXT("[%s]'s dynamic-assets paths:"), *URRTypeUtils::GetERRResourceDataTypeAsString(dataType));
-        for (const auto& path : GetDynamicAssetsPathList(dataType))
+        UE_LOG(LogRapyutaCore,
+               Display,
+               TEXT("[%s]'s dynamic-assets base paths:"),
+               *URRTypeUtils::GetERRResourceDataTypeAsString(dataType));
+        for (const auto& basePath : GetDynamicAssetsBasePathList(dataType))
         {
-            UE_LOG(LogRapyutaCore, Display, TEXT("%s"), *path);
+            UE_LOG(LogRapyutaCore, Display, TEXT("%s"), *basePath);
         }
     }
 }
--- a/Source/RapyutaSimulationPlugins/Private/Core/RRGameSingleton.cpp
+++ b/Source/RapyutaSimulationPlugins/Private/Core/RRGameSingleton.cpp
@@ -30,7 +30,26 @@ URRGameSingleton::~URRGameSingleton()
 
 void URRGameSingleton::PrintSimConfig() const
 {
-    UE_LOG_WITH_INFO(LogRapyutaCore, Display, TEXT("- SIM PROFILING: %d"), BSIM_PROFILING);
+    UE_LOG(LogRapyutaCore, Display, TEXT("RRGameSingleton Configs:"));
+    UE_LOG(LogRapyutaCore, Display, TEXT("- SIM PROFILING: %d"), BSIM_PROFILING);
+    UE_LOG(LogRapyutaCore, Display, TEXT("- ASSETS_RUNTIME_BP_SAVE_BASE_PATH: %s"), *ASSETS_RUNTIME_BP_SAVE_BASE_PATH);
+    UE_LOG(LogRapyutaCore, Display, TEXT("- FOLDER_PATH_ASSET_STATIC_MESHES: %s"), *FOLDER_PATH_ASSET_STATIC_MESHES);
+    UE_LOG(LogRapyutaCore, Display, TEXT("- FOLDER_PATH_ASSET_SKELETAL_MESHES: %s"), *FOLDER_PATH_ASSET_SKELETAL_MESHES);
+    UE_LOG(LogRapyutaCore, Display, TEXT("- FOLDER_PATH_ASSET_SKELETONS: %s"), *FOLDER_PATH_ASSET_SKELETONS);
+    UE_LOG(LogRapyutaCore, Display, TEXT("- FOLDER_PATH_PHYSICS_ASSETS: %s"), *FOLDER_PATH_PHYSICS_ASSETS);
+    UE_LOG(LogRapyutaCore, Display, TEXT("- FOLDER_PATH_ASSET_MATERIALS: %s"), *FOLDER_PATH_ASSET_MATERIALS);
+    UE_LOG(LogRapyutaCore, Display, TEXT("- FOLDER_PATH_ASSET_TEXTURES: %s"), *FOLDER_PATH_ASSET_TEXTURES);
+    UE_LOG(LogRapyutaCore, Display, TEXT("- FOLDER_PATH_ASSET_DATA_TABLES: %s"), *FOLDER_PATH_ASSET_DATA_TABLES);
+    for (auto i = static_cast<int8>(ERRResourceDataType::NONE); i < static_cast<int8>(ERRResourceDataType::TOTAL); ++i)
+    {
+        const ERRResourceDataType dataType = static_cast<ERRResourceDataType>(i);
+        UE_LOG(
+            LogRapyutaCore, Display, TEXT("[%s]'s dynamic-assets paths:"), *URRTypeUtils::GetERRResourceDataTypeAsString(dataType));
+        for (const auto& path : GetDynamicAssetsPathList(dataType))
+        {
+            UE_LOG(LogRapyutaCore, Display, TEXT("%s"), *path);
+        }
+    }
 }
 
 URRGameSingleton* URRGameSingleton::Get()
@@ -62,7 +81,7 @@ URRGameSingleton* URRGameSingleton::Get()
     return singleton;
 }
 
-bool URRGameSingleton::InitializeResources()
+bool URRGameSingleton::InitializeResources(bool bInRequestResourceLoading)
 {
     // Prepare an empty [ResourceMap]
     for (uint8 i = (static_cast<uint8>(ERRResourceDataType::NONE) + 1); i < static_cast<uint8>(ERRResourceDataType::TOTAL); ++i)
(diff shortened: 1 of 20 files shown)


Source/RapyutaSimulationPlugins/Private/Robots/RRBaseRobot.cpp
```diff
@@ -584,61 +584,72 @@ void ARRBaseRobot::Tick(float DeltaSeconds)
...
+    // Verify [ROSSpawnParameters], which house JSON
+    if (nullptr == ROSSpawnParameters)
+    {
+        // ELSE: This could only be a BP robot & Editor operations might be being done
+        RR_VERIFY_STATIC_BP_ROBOT(this);
+        return false;
+    }
+    return true;
+
+#if RAPYUTA_SIM_DEBUG
```
yuokamoto: Isn't this annoying since you will always get log like `[bool_value] not found in json config` with RAPYUTA_SIM_DEBUG==true?

Source/RapyutaSimulationPlugins/Private/Robots/RRBaseRobot.cpp
```diff
@@ -150,7 +150,7 @@ void ARRBaseRobot::PreInitializeComponents()
 {
     if (ROSSpawnParameters)
     {
-        InitPropertiesFromJSON();
+        ensure(InitPropertiesFromJSON());
```
yuokamoto: ensure will stop sim if it uses debugger.
I think it is too much to stop sim with parameter parsing.

Source/RapyutaSimulationPlugins/Private/Robots/RRBaseRobot.cpp
```diff
@@ -584,61 +584,72 @@ void ARRBaseRobot::Tick(float DeltaSeconds)
...
 }
 
-void ARRBaseRobot::InitPropertiesFromJSON()
+bool ARRBaseRobot::InitPropertiesFromJSON()
 {
+    // Verify [ROSSpawnParameters], which house JSON
+    if (nullptr == ROSSpawnParameters)
+    {
+        // ELSE: This could only be a BP robot & Editor operations might be being done
+        RR_VERIFY_STATIC_BP_ROBOT(this);
```
yuokamoto: - currently this case is not happen since this func called inside `if (ROSSpawnParameters)`
-  Doesn't this happen if user want to spawn second robot which is dynamically generated from URDF from inside UE?

yuokamoto:
What is the reason of changes in MeshActor, StaticMesh and procedualMesh? just minor update/fix?

Source/RapyutaSimulationPlugins/Public/Core/RRBaseActor.h
```diff
@@ -112,6 +112,23 @@ class RAPYUTASIMULATIONPLUGINS_API ARRBaseActor : public APawn
...
 public:
+    /**
+     * @brief ACTOR INTIALIZING GENERAL INFO (Unique name, mesh list, material list, etc.)
+     *
+     * @tparam TActorSpawnInfo
+     * @param InActorInfo
+     * @return true
+     * @return false
+     */
+    template<typename TActorSpawnInfo>
```
yuokamoto, line 123: - Do we need to make this template?
- Can we make this merge to Intiialize()? i.e.
```
bool Initialize(const TActorSpawnInfo& InActorInfo = nullptr)
{
 if(InActorInfo){ 
    ActorInfo=InActorInfo
}
...

File: Source/RapyutaSimulationPlugins/Public/Robots/RRRobotStructs.h [hunk 5]
```diff
@@ -1119,31 +1119,76 @@ struct FRRRobotModelData
...
+        {
+            if (jointProp.Name == InJointName)
+            {
+                return jointProp;
+            }
+        }
+        return FRRRobotJointProperty();
+    }
 
+    FRRRobotJointProperty GetJointPropFromChildLinkName(const FString& InChildLinkName) const
```
yuokamoto, line 1181: it is better to add comment that this method return first joint hose childlink name is InChildLinkName.
We may support parallelink robot in the future which have multiple links connects to same link.

Source/RapyutaSimulationPlugins/Public/Core/RRGameSingleton.h
```diff
@@ -225,17 +233,22 @@ class RAPYUTASIMULATIONPLUGINS_API URRGameSingleton : public UObject
...
+     * @note #SASSET_OWNING_MODULE_NAMES must have been initialized earlier & once only at Sim start by [ConfigureSimInPlay()]
      * @param InDataType
      * @return TArray<FString>
      */
-    FORCEINLINE static TArray<FString> GetDynamicAssetsPathList(const ERRResourceDataType InDataType)
+    FORCEINLINE static TArray<FString> GetDynamicAssetsBasePathList(const ERRResourceDataType InDataType)
     {
-        static TArray<FString> runtimeAssetsPathList;
-        for (const auto& moduleName : SASSET_OWNING_MODULE_NAMES[InDataType])
+        static TArray<FString> dynamicAssetsBasePathList = [InDataType]()
```
yuokamoto, line 242: what is the main reason to make this lambda function?

Source/RapyutaSimulationPlugins/Private/Core/RRGameSingleton.cpp
```diff
@@ -62,7 +83,7 @@ URRGameSingleton* URRGameSingleton::Get()
     return singleton;
 }
 
-bool URRGameSingleton::InitializeResources()
+bool URRGameSingleton::InitializeResources(bool bInRequestResourceLoading)
```
yuokamoto, line 86: When you use this with bInRequestResourceLoading=false ?

Source/RapyutaSimulationPlugins/Private/Robots/RRBaseRobot.cpp
Tadinu: Yeah since commenting out codes should be discouraged, and by convention we put it into debug directive to make it formally debugged code, but in this case, it is more of annotation so I suppose better move to doxygen

Tadinu: Yeah since commenting out codes should be discouraged, and by convention, we put it into debug directive to be formally denoted as debug code, but in this case, it is more of annotation so I suppose better move to doxygen

Source/RapyutaSimulationPlugins/Private/Robots/RRBaseRobot.cpp
Tadinu: oh really, I suppose ensure only print the callstack & only verify/check will stop sim, ok so here not so essential since error is already logged in the function

Source/RapyutaSimulationPlugins/Private/Robots/RRBaseRobot.cpp
Tadinu: Ah no, all robot instances go through the same flow of calls here, so right though it is already checked by the caller  (as in `ARRBaseRobot::PreInitializeComponents()`), but that's the caller logics only, which this function is not aware of:
* So ideally the check for `ROSSpawnParameters` should be always put inside this function due to its role as providing Json param inputs.
-> This is more robust that way, when `InitPropertiesFromJSON()` will happen to be moved somewhere else.
* Previously, we hadn't made a bool return, but now we have so I figure adding a basic check serves its purpose in `ARRBaseRobot` as the top parent function, which always needs to return `true` before executing the child overriding ones.

Tadinu: Ah no, all robot instances go through the same flow of calls here, so right though it is already checked by the caller  (as in `ARRBaseRobot::PreInitializeComponents()`), but that's the caller logics only, which this function is not aware of:
* So ideally the check for `ROSSpawnParameters` should be always put inside this function due to its role as providing Json param inputs.
-> It is more robust that way, when `InitPropertiesFromJSON()` will happen to be moved somewhere else.
* Previously, we hadn't made a bool return, but now we have so I figure adding a basic check serves its purpose in `ARRBaseRobot` as the top parent function, which always needs to return `true` before executing the child overriding ones.

yuokamoto: Are there any case that robot from URDF with ROSSpawnParameters == nullptr?

Source/RapyutaSimulationPlugins/Public/Core/RRBaseActor.h
Tadinu: So the template's purpose is to allow users to use a custom ActorSpawnInfo struct inheriting from `FRRActorSpawnInfo` and still be able to instantiate it.
Right, merging them into one makes sense.

Source/RapyutaSimulationPlugins/Private/Robots/RRBaseRobot.cpp
Tadinu: So, due to essentially instantiated in a ROS-spawned robot only, if sb (like UE devs) spawns it independently in UE (using UE pure spawning) from URDF, yes it will be null.
Also, it is null in case of BP robot as being opened in the Editor

Source/RapyutaSimulationPlugins/Public/Core/RRBaseActor.h
Tadinu: So the template's purpose is to allow users to use a custom ActorSpawnInfo struct inheriting from `FRRActorSpawnInfo` and still be able to instantiate it.
Ah I forgot, since a template cannot be made `virtual`, which is required here so yeah we could not merge those for that reason.

Source/RapyutaSimulationPlugins/Private/Robots/RRBaseRobot.cpp
yuokamoto: I think, then we don't need to put `RR_VERIFY_STATIC_BP_ROBOT` here which print error log since it is not error.

Source/RapyutaSimulationPlugins/Private/Robots/RRBaseRobot.cpp
yuokamoto: then please add this
https://www.doxygen.nl/manual/commands.html#cmdcode

Source/RapyutaSimulationPlugins/Private/Robots/RRBaseRobot.cpp
Tadinu: Ah right, to be rem

Source/RapyutaSimulationPlugins/Public/Core/RRGameSingleton.h
Tadinu: That is for you to initialize the value once only as the return of an inline lambda call, due to `SASSET_OWNING_MODULE_NAMES` being defined once only at Sim start

Source/RapyutaSimulationPlugins/Private/Core/RRGameSingleton.cpp
Tadinu: As noted in the description (pls check it first), it is an option for users to avoid early loading `uasset` in DynamicContents if there is no need for it (eg in a pure BP-based application)

yuokamoto: My point is that there is no use case for it yet but for future use case?

Tadinu: ah yes,  It is kept here for the mentioned use case, with the purpose to make RRGameSingleton less coupling. In cmdlet I used to experiment with it but still ended up requiring runtime resources.

yuokamoto (approved):

Source/RapyutaSimulationPlugins/Private/Core/RRStaticMeshComponent.cpp
```diff
@@ -65,13 +65,15 @@ void URRStaticMeshComponent::SetMesh(UStaticMesh* InStaticMesh)
...
-            verify(bodySetup->bCreatedPhysicsMeshes);
-            verify(false == bodySetup->bFailedToCreatePhysicsMeshes);
-            verify(bodySetup->bHasCookedCollisionData);
+            ensure(EBodyCollisionResponse::BodyCollision_Enabled == bodySetup->CollisionReponse);
+            if (bUseDefaultSimpleCollision)
+            {
+                ensure(bodySetup->bCreatedPhysicsMeshes);
+                ensure(false == bodySetup->bFailedToCreatePhysicsMeshes);
+                ensure(bodySetup->bHasCookedCollisionData);
+            }
```
Tadinu, line 76: verify -> ensure to avoid crash top in PIE/Standalone/package

Source/RapyutaSimulationPlugins/Public/Core/RRMeshActor.h
```diff
@@ -36,42 +36,25 @@ class RAPYUTASIMULATIONPLUGINS_API ARRMeshActor : public ARRBaseActor
...
-
-    /**
-     * @brief Initialize MeshActor by 1)Create child mesh components, 2)#SetCustomDepthEnabled, 3)Set mobility. 
-     * 
-     * @return true 
-     * @return false 
+     * @brief Initialize MeshActor by 1)Create child mesh components, 2)#SetCustomDepthEnabled, 3)Set mobility.
+     *
+     * @return true
+     * @return false
```
Tadinu, line 42: `InitializeWithSpawnInfo()` moved to `ARRBaseActor` to be invokable also by non-template version of `SpawnSimActor()`

PR #178, opened 2023-06-26
ERRRobotDescriptionType add SINGLE_CAD

Tadinu:
SINGLE_CAD is a description type whereby an entity (robot or object) is imported purely from a single 3D CAD file. In case of robots, only FBX is supported for now.
`ARRBaseRobot` `RobotModelName` to ref `EntityModelName` as both referring to RRBaseActor's model name. It is just kept so to make it clear for robots.

### DIFF
--- a/Source/RapyutaSimulationPlugins/Public/Core/RRActorCommon.h
+++ b/Source/RapyutaSimulationPlugins/Public/Core/RRActorCommon.h
@@ -29,6 +29,7 @@
 
 #define RAPYUTA_SIM_VERBOSE (0)    // todo make this CVar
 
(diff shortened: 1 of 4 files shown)


yuokamoto:
out of this pr but can we apply same things for RobotVehicleMoveCompoment and MoveComponent?

yuokamoto (approved):

Tadinu:
So the reference only works for the same type members (like an alias, there is always 1 instance of it in memory), thus could not be applied to Movement component ones due to their being of, though parent-child pointer types, but still different types.
In fact, MovementComponent, eg in case of being instantiated as a ChaosVehicleDriveComp, also effectively means RobotVehicleMoveCompoment is always null.

PR #180, opened 2023-07-01
Revert RobotModelName reference to avoid crash in child BP

yuokamoto:
@Tadinu I experienced tb crash when I opened tht Turtlebot3WaffleVehicle in Editor.
Please check next week


Tadinu (approved):
So it seems BP as typically has issue with native C++, we thus may have to use only EntityModelName.

Tadinu (approved):

PR #182, opened 2023-07-05
Add pak loader

Tadinu:
* `ERRFileType` add PAK
* Add `URRPakLoader`, loading pak files (compressed cooked assets) from `/Game/RapyutaContents/DynamicContents/Paks`, only used in packaged build
* Move EMPTY_STR from ActorCommon to RRGeneralUtils.h to be more decoupling
* Move `ERRFileType & ERRShapeType` from ActorCommon -> ObjectCommon as usable by any generic object
* `URRAssetUtils::SavePackageToAsset()` clear `RF_Transient`, which is not allowed for an object being saved to uasset
* `URRCoreUtils::LoadImageToTexture()` add `bInSaveToAsset`, using `ImportObject()` to import texture if to be saved to uasset, since `FImageUtils::ImportFileAsTexture2D()` only creates a transient non-source-data texture2d that is ineligible to be saved to uasset.

### DIFF
--- a/Source/RapyutaSimulationPlugins/Private/Core/RRCoreUtils.cpp
+++ b/Source/RapyutaSimulationPlugins/Private/Core/RRCoreUtils.cpp
@@ -270,7 +270,7 @@ bool URRCoreUtils::WaitUntilThenAct(TFunctionRef<bool()> InCond,
         // Sleep takes seconds, not msec
         FPlatformProcess::Sleep(InIntervalTimeInSec);
     }
-    // Either InCond() is met or [ElapsedTime] is over [InTimeoutInSec]
+    // Now, either InCond() has been met or [GetElapsedTimeSecs()] is over [InTimeoutInSec]
 
     if (bResult)
     {
--- a/Source/RapyutaSimulationPlugins/Private/Core/RRCoreUtils.cpp
+++ b/Source/RapyutaSimulationPlugins/Private/Core/RRCoreUtils.cpp
@@ -7,9 +7,13 @@
 #include "HAL/PlatformProcess.h"
 #include "IESConverter.h"
 #include "ImageUtils.h"
+#if WITH_EDITOR
+#include "Factories/TextureFactory.h"
+#endif
 
 // RapyutaSimulationPlugins
 #include "Core/RRActorCommon.h"
+#include "Core/RRAssetUtils.h"
 #include "Core/RRBaseActor.h"
 #include "Core/RRGameMode.h"
 #include "Core/RRGameSingleton.h"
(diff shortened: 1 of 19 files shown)


Source/RapyutaSimulationPlugins/Private/Core/RRGameSingleton.cpp
```diff
@@ -97,6 +100,33 @@ bool URRGameSingleton::InitializeResources(bool bInRequestResourceLoading)
...
         // READ ALL SIM DYNAMIC RESOURCES (UASSETS) INFO FROM DESGINATED [~CONTENT] FOLDERS
         // & REGISTER THEM TO BE ASYNC LOADED INTO [ResourceMap]
+#if (!WITH_EDITOR)
+        // [PAK] --
+        // NOTE: PakLoader is only available in packaged Sim
+        PakLoader = URRUObjectUtils::CreateSelfSubobject<URRPakLoader>(this, TEXT("RRPakLoader"));
+        if (PakLoader->Initialize())
+        {
+            // Load pak files
+            for (const auto& paksBasePath : GetDynamicAssetsBasePathList(ERRResourceDataType::UE_PAK))
```
yuokamoto, line 110: Do we need to load packaged asset?
It causes load all asset in the begining and make start up time slower and slower

I think this is fine as a first step but should stat thinking about loading as demands. we maybe define list of initial loading asset DefaultConfig.ini

yuokamoto (commented):
Do we need both `UE_PAK_MOUNTED_BASE_FOLDER_NAME` and `FOLDER_PATH_ASSET_PAKS` ?

Source/RapyutaSimulationPlugins/Private/Core/RRGameSingleton.cpp
Tadinu: So this is already added as backlog task to be done.
Basically Sim should load only resources of entity models specified by users, but we need to define configs for them, similar to Import~ModelNameList arg of ImportEntitiesCmdlet.
When users define it, probably in INI, only resource assets (uncooked and Paks) that correspond to stipulated models are loaded.

Source/RapyutaSimulationPlugins/Public/Core/RRCoreUtils.h
```diff
@@ -613,6 +617,31 @@ class RAPYUTASIMULATIONPLUGINS_API URRCoreUtils : public UBlueprintFunctionLibra
...
+        }
+
+        // Wait for it to finish with [InTimeOutSecs]
+        const float lastMarkedTime = URRCoreUtils::GetSeconds();
+        while (InProcess->Update())
+        {
+            // Already slept in [Update()]
+            if (URRCoreUtils::GetElapsedTimeSecs(lastMarkedTime) > InTimeOutSecs)
+            {
+                InProcess->Cancel();
```
yuokamoto, line 648: Don't you need break while loop or Update() will return false?

Tadinu: Yeah I thought about it but then figured out the api actually already handles it more properly to make sure the while loop here only breaks after the process is terminated, specifically
* [Cancel() ](https://github.com/EpicGames/UnrealEngine/blob/release/Engine/Source/Runtime/Core/Public/Misc/MonitoredProcess.h#L71)-> ON `Canceling`
* [FMonitoredProcess::TickInternal()](https://github.com/EpicGames/UnrealEngine/blob/release/Engine/Source/Runtime/Core/Private/Misc/MonitoredProcess.cpp#L135) -> upon cancelling turns OFF `bIsRunning` (which is of atomic type thus thread-safe) **only after officially terminating the process**, also making our while loop here break accordingly

Tadinu:
yeah UE_PAK_MOUNTED_BASE_FOLDER_NAME was only for debug, removed

Source/RapyutaSimulationPlugins/Private/Core/RRCoreUtils.cpp
```diff
@@ -261,18 +261,16 @@ bool URRCoreUtils::WaitUntilThenAct(TFunctionRef<bool()> InCond,
...
-    while (!bResult && elapsed_time < InTimeoutInSec)
+    while (!bResult && (URRCoreUtils::GetElapsedTimeSecs(startTime) < InTimeoutInSec))
     {
         bResult = InCond();
         // Sleep takes seconds, not msec
         FPlatformProcess::Sleep(InIntervalTimeInSec);
-        elapsed_time = FTimespan(FDateTime::UtcNow() - begin).GetTotalSeconds();
     }
-    // Either InCond() is met or [elapsed_ticks] is over [InTimeoutInSec]
+    // Either InCond() is met or [ElapsedTime] is over [InTimeoutInSec]
```
yuokamoto: [ElapsedTime] looks like variable

yuokamoto (approved):

PR #183, opened 2023-07-11
Fetch BP classes in packaged build

Tadinu:
`URRAssetUtils::FindBlueprintClass()` moved to cpp, In pkg build fetch bp generated class instead of bp itself as in Editor
`URRPakLoader` also scan runtime BP save assets path after mounting PAKs, which could also contain packed BPs.

### DIFF
--- a/Source/RapyutaSimulationPlugins/Private/Core/RRAssetUtils.cpp
+++ b/Source/RapyutaSimulationPlugins/Private/Core/RRAssetUtils.cpp
@@ -18,6 +18,69 @@
 #include "Core/RRGameSingleton.h"
 #include "Core/RRThreadUtils.h"
 
+using URRBlueprint = typename TChooseClass<WITH_EDITOR, UBlueprint, UBlueprintGeneratedClass>::Result;
+
+UClass* URRAssetUtils::FindBlueprintClass(const FString& InBlueprintClassName)
+{
+    IAssetRegistry& assetRegistry = GetAssetRegistry();
+    if (assetRegistry.IsLoadingAssets())
+    {
+        assetRegistry.SearchAllAssets(true);
+    }
+
+#if WITH_EDITOR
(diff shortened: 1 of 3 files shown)


Source/RapyutaSimulationPlugins/Private/Core/RRAssetUtils.cpp
```diff
@@ -18,6 +18,69 @@
...
+
+UClass* URRAssetUtils::FindBlueprintClass(const FString& InBlueprintClassName)
+{
+    IAssetRegistry& assetRegistry = GetAssetRegistry();
+    if (assetRegistry.IsLoadingAssets())
+    {
+        assetRegistry.SearchAllAssets(true);
+    }
+
+#if WITH_EDITOR
```
yuokamoto, line 31: Just curous, this measn target BP name is different in WITH_EDITOR or not?

Tadinu: So as observed (also noted in this [link](https://github.com/rapyuta-robotics/RapyutaSimulationPlugins/pull/183/files#diff-91b1460caf2c09463daf0eec897edc102002be9327eb1d492dad52fe4f3ce384R384)), basically for a BP .uasset:
* In Editor, asset registry only register its `UBlueprint` class, from which the very BP class itself must be fetched from `UBlueprint::GeneratedClass`
* In packaged build, asset registry instead register directly its `UBlueprintGeneratedClass` as for optimization I suppose, due to no need for UBlueprint info.
-> As in the comment, only the latter has `_C` trailing in its asset name.

yuokamoto: Got it. Thank you.

yuokamoto (approved):

PR #197, opened 2023-08-16
Robot ->Entity model data structs + Only load resources upon access

Tadinu:
Toward the unification of robot & object struct definitions & various models-related handling, this PR:
- Replaces `Robot` with `Entity` for entity-generic struct names (except for certain robots-specific-worth ones like links/joints)
- Is a bit invasive refactoring, but worth it for the sake of avoiding confusion over whether or not the model treatment is applied collectively for robots & objects, also reducing lots of api duplication among robot, object, & entity world accordingly.
- Most critically, only load entity resource on request, instead of loading all of them initially

Specific notable changes include:
* `URRGameSingleton::InitializeResources()` `bInRequestResourceLoading` dft False, only collating metadata, except for UE_DATA_TABLE
* `FRRRobotModelInfo -> FRREntityModelInfo`, `RRRobotStructs.h -> RREntityStructs.h` & various robot->entity prefixed-types


Source/RapyutaSimulationPlugins/Public/Core/RRMeshData.h
```diff
@@ -62,6 +63,53 @@ struct RAPYUTASIMULATIONPLUGINS_API FRRBoneProperty
     UPROPERTY(VisibleAnywhere)
     float Mass = 0.f;
 
+    FRRRobotLinkProperty ConvertToLinkProperty() const
```
yuokamoto, line 66: what is the difference between FRRRobotLinkProperty and FRRBoneProperty?

Tadinu: So as we know, bone is the concept used in UE skeleton, skeletal mesh
https://github.com/EpicGames/UnrealEngine/blob/release/Engine/Source/Runtime/AnimationCore/Public/BoneIndices.h
Robot link is mostly a data representation of robot description format like URDF/SDF
* When we build skeletal mesh, skeleton, we need to construct the bone structure from robot link info loaded from URDF/SDF
* However, if there is no URDF/SDF as in case of loading from a single CAD (like FBX), this conversion util is used to auto-construct a template robot model info from bone info, loaded from FBX.

Source/RapyutaSimulationPlugins/Private/Core/RRGameSingleton.cpp
```diff
@@ -160,16 +166,74 @@ bool URRGameSingleton::InitializeResources(bool bInRequestResourceLoading)
...
+        bResult &= RequestResourcesLoading<ERRResourceDataType::UE_DATA_TABLE>();
+
+        // -> Only collate assets' metadata without loading
+        CollateAssetsInfo();
+
+        // NOTE: Due to on-the-fly resource loading is only possible if running in Game-thread
+        // -> Preload global statically defined resources, in case they are referenced in non-game thread
+        // -> Could only load after collating assets info above
+        // StaticMeshes
+        GetStaticMesh(SHAPE_NAME_PLANE);
```
chalice-graeme, line 187: Are these resources guaranteed to exist? In a packaged game, Unreal _should_ exclude anything that is not explicitly referenced in the scene hierarchy it generates to build all the packages.

Source/RapyutaSimulationPlugins/Private/Core/RRUObjectUtils.cpp
```diff
@@ -477,7 +477,7 @@ void URRUObjectUtils::ApplyMaterialProps(UMaterialInstanceDynamic* InMaterial,
         {
             InMaterial->SetTextureParameterValue(
                 FRRMaterialProperty::PROP_NAME_ALBEDO,
-                gameSingleton->GetTexture(URRMathUtils::GetRandomElement(InMaterialInfo.AlbedoTextureNameList)));
+                gameSingleton->GetTexture(URRMathUtils::GetRandomElement(InMaterialInfo.AlbedoTextureNameList), false));
```
chalice-graeme, line 480: What is the purpose of the GetRandomElement? Is it just for testing?

Source/RapyutaSimulationPlugins/Private/Core/RRGameSingleton.cpp
Tadinu: So `/RapyutaSimulationPlugins/DynamicContents` is expected to be added to `DirectoriesToAlwaysCook` in `DefaultGame.ini` like [here](https://github.com/rapyuta-robotics/flaptter_UE/blob/devel/Config/DefaultGame.ini#L86)
Otherwise, an error will be raised.
Since these are commonly used basic primitive shapes and master material used in our Sim, I keep them fetched here for ease of managing.
Regarding `DirectoriesToAlwaysCook`, I will add further guideline logs into the error, if user project does not have it configured.

Source/RapyutaSimulationPlugins/Private/Core/RRUObjectUtils.cpp
Tadinu: So for some object like `pallet`, it could be configured with various albedo textures that are stored as `AlbedoTextureNameList`. For robot sim, it does not matter which texture is being used for now, for synthetic data sim, it is required to be randomly picked.

yuokamoto (approved):

chalice-graeme (approved):

PR #207, opened 2023-09-12
RRCoreUtils add RenderThumbnail() & GenerateThumbnail()

Tadinu:
RRStaticMeshComponent Gen visualMesh's thumbnail also as saving to uasset
NOTE: This is not supported as running in Standalone or packaged build, only possible in Editor or cmdlet.

### DIFF
--- a/Source/RapyutaSimulationPlugins/Private/Core/RRCoreUtils.cpp
+++ b/Source/RapyutaSimulationPlugins/Private/Core/RRCoreUtils.cpp
@@ -538,7 +538,7 @@ bool URRCoreUtils::RenderThumbnail(UObject* InObject,
         renderTarget->RenderTargetFormat = RTF_RGBA8;
         return renderTarget;
     }();
-    sThumbnailRenderTarget->InitCustomFormat(InImageWidth, InImageHeight, PF_B8G8R8A8, false);
+    sThumbnailRenderTarget->InitAutoFormat(InImageWidth, InImageHeight);
     auto* renderTargetResource = sThumbnailRenderTarget->GameThread_GetRenderTargetResource();
 
     // Create a canvas for the render target and clear it to black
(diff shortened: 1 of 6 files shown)


yuokamoto (approved):
lgtm

Source/RapyutaSimulationPlugins/Private/Core/RRCoreUtils.cpp
```diff
@@ -478,3 +480,173 @@ bool URRCoreUtils::LoadIESProfilesFromFolder(const FString& InFolderPath,
...
+    {
+        // Create thumbnail render target
+        UTextureRenderTarget2D* renderTarget = NewObject<UTextureRenderTarget2D>(GetTransientPackage());
+        renderTarget->AddToRoot();
+        renderTarget->ClearColor = FLinearColor::Transparent;
+        renderTarget->TargetGamma = GEngine->GetDisplayGamma();
+        renderTarget->RenderTargetFormat = RTF_RGBA8;
+        return renderTarget;
+    }();
+    sThumbnailRenderTarget->InitCustomFormat(InImageWidth, InImageHeight, PF_B8G8R8A8, false);
```
chalice-graeme: The pixel format here does not match the render target format a few lines above. I suppose this was tested and works okay, but you might check that there isn't an RTF_BGRA8 for the above line. (For what it's worth, in modern graphics APIs, BRGA is the only guaranteed-to-be-supported format for render targets.)

chalice-graeme (approved):

Source/RapyutaSimulationPlugins/Private/Core/RRCoreUtils.cpp
Tadinu: Right, it looks like `ETextureRenderTargetFormat` is UE-defined only for render target, and there is a specific mapping to the pixel format here
https://github.com/EpicGames/UnrealEngine/blob/release/Engine/Source/Runtime/Engine/Classes/Engine/TextureRenderTarget2D.h#L52
So the latter setting is redundant, will remove it.
Also, I take refs from here
https://github.com/EpicGames/UnrealEngine/blob/release/Engine/Source/Editor/UnrealEd/Private/ObjectTools.cpp#L5057
https://github.com/EpicGames/UnrealEngine/blob/release/Engine/Source/Editor/UnrealEd/Private/EditorEngine.cpp#L2766

PR #272, opened 2024-03-16
Refactor AI agent with Behavior tree

yuokamoto:
image (10)
image (9)
image (8)
image (7)


Source/RapyutaSimulationPlugins/Private/Robots/RRAIRobotROSController.cpp
```diff
@@ -1095,13 +1100,48 @@ FTransform ARRAIRobotROSController::GetOrigin()
...
+}
+
 void ARRAIRobotROSController::ModeUpdate()
 {
     if (Mode != ERRAIRobotMode::MANUAL)
     {
         if (NavStatus == ERRAIRobotNavStatus::IDLE)
         {
             FMoveCompleteCallback onSuccess, onFail;    // dummy
+            FTransform dummy;
```
tetorea, line 1144: It seems dummy is initialized after but never used

yuokamoto: it is used  at line 1155 and 1158

tetorea: these lines initialize dummy but then it is not used?

yuokamoto: Those are mainly used by Blueprint to get output transform directly.

tetorea (approved):

PR #278, opened 2024-05-05
update physics param of tb to avoid drifting and non linear movement

yuokamoto:
Change physics joint's damper value, i.e. strength value.

original
image

new
image


yuokamoto:
added with different PR

PR #307, opened 2024-10-03
update to make it useful for children class

yuokamoto:


Source/RapyutaSimulationPlugins/Private/Sensors/RR3DLidarComponent.cpp
```diff
@@ -388,7 +388,7 @@ FROSPointCloud2 URR3DLidarComponent::GetROS2Data()
             }
 
             // Convert pose to local coordinate, ROS unit and double -> float
-            FVector posInDouble = RecordedHits.Last(index).ImpactPoint + BWithNoise * PositionNoise->Get();
+            FVector posInDouble = RecordedHits.Last(index).ImpactPoint;    // + BWithNoise * PositionNoise->Get();
```
tetorea, line 391: why is the noise removed?
so that it can be added in child class?

tetorea (approved):

Source/RapyutaSimulationPlugins/Private/Sensors/RR3DLidarComponent.cpp
yuokamoto: Thank you.
I found noise is added here already
https://github.com/rapyuta-robotics/RapyutaSimulationPlugins/blob/9dce510b48a16122855d48c18032741708a305d3/Source/RapyutaSimulationPlugins/Private/Sensors/RR3DLidarComponent.cpp#L145

so I think it is added twice originally

arturocuma (approved):

PR #323, opened 2024-11-06
Fix imu frameid

yuokamoto:


prodrigues-rapyuta:
I found no more issues on the fix
The simulation is working properly, now with the correct frame_id

PR #329, opened 2024-11-14
fix (RRROS2IMUComponent.cpp): make gravity vector of IMU follow standard

prodrigues-rapyuta:
The expected signal for the gravity vector when read through IMU is upwards, such as the example image

image

I think it should be better to keep the negative sign as default.

Source:[Why does an accelerometer measure gravity with positive sign?](https://base.movella.com/s/article/Why-does-an-accelerometer-measure-gravity-with-positive-sign?language=en_US)


yuokamoto (approved):
Thank you.
Please add url as a comment in the code, then we can easy find why in the future.

yuokamoto (approved):

PR #336, opened 2024-11-24
feat: add conditional rendering in scene capture logic

bresilla:
A very small conditional rendering "check" so can be used in Blueprints to render or not render scene. When not used can be switched off so we don't need to do rendering calculations. 

- Add a `Render` bool property to control scene capture execution.
- Wrap the scene capture logic in a conditional check on `Render` to allow for selective rendering.


yuokamoto (approved):

yuokamoto:
Thank you!

PR #345, opened 2024-12-02
Improve RTF for 3D Livox Lidar 

arturocuma:
Comment traceHandle on ASYNC for now, not sure if this should be removed

### DIFF
--- a/Source/RapyutaSimulationPlugins/Private/Sensors/RR3DLidarComponent.cpp
+++ b/Source/RapyutaSimulationPlugins/Private/Sensors/RR3DLidarComponent.cpp
@@ -14,33 +14,33 @@ void URR3DLidarComponent::TickComponent(float DeltaTime, enum ELevelTick TickTyp
 {
     Super::TickComponent(DeltaTime, TickType, ThisTickFunction);
 #if TRACE_ASYNC
-    verify(TraceHandles.Num() == RecordedHits.Num());
-    UWorld* world = GetWorld();
-    for (auto i = 0; i < TraceHandles.Num(); ++i)
-    {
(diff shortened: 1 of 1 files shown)


yuokamoto (commented):
you should not just comment out but override in child class

arturocuma:
> you should not just comment out but override in child class

@yuokamoto it is already overwritten but it calls Super::TickComponent(DeltaTime, TickType, ThisTickFunction);
Should I inherit directly RRLiVoxMid360Component from URRBaseLidarComponent?

yuokamoto:
> > you should not just comment out but override in child class
> 
> @yuokamoto it is already overwritten but it calls Super::TickComponent(DeltaTime, TickType, ThisTickFunction); Should I inherit directly RRLiVoxMid360Component from URRBaseLidarComponent?

1. if most of the logic is not same, you can inherit from URRBaseLidarComponent
2. you can change it not call super Tick or call URRBaseLidarComponent::TickComponent instead of Super::TickComponent

arturocuma:
> > > you should not just comment out but override in child class
> > 
> > 
> > @yuokamoto it is already overwritten but it calls Super::TickComponent(DeltaTime, TickType, ThisTickFunction); Should I inherit directly RRLiVoxMid360Component from URRBaseLidarComponent?
> 
> 1. if most of the logic is not same, you can inherit from URRBaseLidarComponent
> 2. you can change it not call super Tick or call URRBaseLidarComponent::TickComponent instead of Super::TickComponent

Thanks, I will go with call URRBaseLidarComponent::TickComponent instead of Super::TickComponent

arturocuma:
We can close this PR as the changes can be done on RapyutaSimulationInternal

PR #364, opened 2025-04-19
fix: packaging error in shipping mode

james-yoo:
This PR fixes an error that prevented the 'IsDebugEnabled' method from being referenced when packaging in shipping mode.


yuokamoto (approved):
Thanks always

